PR #5, opened 2020-09-29
Add sentry

harshjv:


monokh:
I think we were not conclusive about adding sentry, can raise again when we are clear

PR #10, opened 2020-12-02
Fix qa changes

bradleySuira:


monokh:
The code is fine, but i've shared some design issues in slack

PR #12, opened 2020-12-07
Feature unlock UI changes

bradleySuira:


src/views/Receive.vue
```diff
@@ -58,12 +67,13 @@ export default {
...
     chainName () {
-      const chain = getChainFromAsset(this.asset)
       return ({
         BTC: 'bitcoin',
         ETH: 'ethereum',
-        RBTC: 'ethereum'
-      })[chain]
+        RBTC: 'rbtc/rsk',
+        DAI: 'dai/ethereum',
+        UNI: 'uni/ethereum'
```
monokh, line 75: The chain name is also used as part of the URI for QR code, that has to be "etheruem" for ethereum addresses and "bitcoin" for bitcoin ones. Otherwise the wallets will not be able to detect them correctly.

So i guess this part has to be moved to another computed specifically for displaying the message.

Also this won't be working for other ERC20 assets. I'm seeing it as "YOUR CURRENT ADDRESS"

PR #15, opened 2020-12-10
Feature qa comments fixes

bradleySuira:


src/assets/scss/style.scss
```diff
@@ -202,8 +204,100 @@ textarea {
   font-weight: $font-weight-base;
 }
 
+// overrided btn-outline and primary as design spect for secondary buttons
+.btn-primary {
+  color: #ffffff;
```
monokh, line 209: Do you know if there is any way in bootstrap to customize this so that we don't have to be specific to each button? 
I guess design wants something very custom

bradleySuira: I tried to use the same mixins or overriding variables but yes the requirements are very specific, the other option is tu use or own button prefix like liq-primary-btn, etc

bradleySuira: Screen Shot 2020-12-11 at 10 33 34 AM

PR #18, opened 2020-12-11
Feature activity view

bradleySuira:


src/router/index.js
```diff
@@ -69,24 +72,50 @@ const routes = [
...
+    ]
+  },
+  // Details
+  {
+    path: '/details',
+    component: Details,
+    props: true,
+    children: [
+      {
+        path: ':id/swap',
```
monokh, line 100: would've thought it should be /details/swap/:id instead

src/store/getters.js
```diff
@@ -53,5 +53,40 @@ export default {
...
+      .sort(([assetA], [assetB]) => {
+        return assets.indexOf(assetA) - assets.indexOf(assetB)
+      })
+  },
+  networkWalletBalances (state) {
+    const { balances, activeNetwork, activeWalletId } = state
+    if (!balances[activeNetwork]) return false
+    if (!balances[activeNetwork][activeWalletId]) return false
+
+    return balances[activeNetwork][activeWalletId]
```
monokh, line 76: something i found recently. you can remove the checks for undefined and instead do
`return balances[activeNetwork]?.[activeWalletId]`

src/utils/history.js
```diff
@@ -59,3 +59,71 @@ export function getStep (item) {
...
+    label: 'Send',
+    icon: 'send'
+  },
+  RECEIVE: {
+    label: 'Receive',
+    icon: 'receive'
+  }
+}
+
+export const ACTIVITY_FILTER_STATUSES = {
```
monokh, line 78: I think the "FILTER" part in these names is not needed? They are used for multiple purposes and the content of the object is not specific to filtering

src/views/Details/Details.vue
```diff
@@ -0,0 +1,59 @@
...
+import { mapGetters } from 'vuex'
+import NavBar from '@/components/NavBar.vue'
+
+export default {
+  components: {
+    NavBar
+  },
+  data () {
+    return {
+      advanced: false,
```
monokh, line 20: I'm confused as to what's going on here
this data is not used in the template
and why is there a router-view inside? Can't this page just be for it's own path?

PR #21, opened 2020-12-21
allow argument of type buffer for sendTx method

harshjv:


monokh:
I believe we found that this usecase can be accommodated using `signPSBT`

PR #22, opened 2020-12-22
Feature ui fixes and swap screens

bradleySuira:


src/components/FeeSelector.vue
```diff
@@ -3,7 +3,10 @@
     <label class="btn btn-option btn-option-lg"
         v-for="name in ['slow', 'average', 'fast']" :key="name"
         :class="{ active: (name === value)}"
-        v-tooltip="{ content: getTooltip(name)} "
+        v-tooltip="{
+          content: getTooltip(name),
+          trigger: txTypes ? 'hover' : 'manual'
```
monokh, line 8: this would mean there is no tooltip displayed I guess? 

in stead can we do this? 

at `if (this.txTypes) {`
We can do `else` that displays `this.fees[name].fee`.
for example `55 gwei`

src/views/Swap.vue
```diff
@@ -455,7 +451,7 @@ export default {
       return this.networkMarketData[this.asset]
     },
     ethRequired () {
-      return this.networkWalletBalances.ETH === 0
+      return this.assetChain === 'ETH' && this.networkWalletBalances.ETH === 0
```
monokh, line 454: `assetChain` would be only the from asset right?
If to asset is also on eth network , you need ETH too.
so i think it should be `[this.assetChain, this.toAssetChain].includes('ETH') && this.networkWalletBalances.ETH === 0`

src/views/Swap.vue
```diff
@@ -552,10 +548,10 @@ export default {
       return this.sendFeeType === FEE_TYPES.BTC
     },
     amountToSendInFiat () {
-      return prettyFiatBalance(this.amount, this.fiatRates[this.asset])
+      return prettyFiatBalance(this.amount, this.fiatRates[this.assetChain])
```
monokh, line 551: I think you may be confusing assetChain and asset 😄  
I will explain externally

src/components/FeeSelector.vue
bradleySuira: fixed

src/views/Swap.vue
bradleySuira: Fixed now, thanks for explain

Issue #39, opened 2021-01-21
If we click on Swap button on a asset that doesn't have swap, appear a blank window


bitcoinuser:
If we click on Swap button on a asset that doesn't have swap, appear a blank window. 
Obs. I think swap button should be disabled on assets that doesn't support swap, or if the button be enabled, when clicked, appear some message that the specific asset has no swap support yet.

![Captura de tela de 2021-01-20 21-29-01](https://user-images.githubusercontent.com/1178856/105257560-ca138200-5b66-11eb-834b-6212c9bcd1f9.png)

Issue #33, opened 2021-01-11
DoC balance do not show correctly (Aligned)


bitcoinuser:
See:

![Captura de tela de 2021-01-11 12-51-58](https://user-images.githubusercontent.com/1178856/104204686-4d690100-540c-11eb-8b3e-746db0acd084.png)

bitcoinuser:
Fixed

PR #51, opened 2021-02-01
Fix prod bugs

bradleySuira:
- Fix for total amount in USD in the send screen review
- Fix for rsk bug from github https://github.com/liquality/wallet/issues/33 (now the size of the font is reduced depending of the amount)
- Fix for rsk bug from github https://github.com/liquality/wallet/issues/39 ,  this issue is related only to rsk because we don't have market data, for now I added validations to show 0 as value and the user can't make the swap also I disabled the inputs because we don't have a way to convert fiat to crypto or crypto to fiat and set the amount to receive


src/utils/coinFormatter.js
```diff
@@ -16,26 +16,26 @@ export const dpUI = (amount, dp = VALUE_DECIMALS) => {
...
+  if (!amount || !coin) return amount
 
   amount = cryptoassets[coin].unitToCurrency(amount)
 
   return dpUI(amount, dp)
 }
 
 export const prettyFiatBalance = (amount, rate) => {
-  if (!amount) return amount
+  if (!amount || !rate) return amount
```
monokh, line 27: If the rate is not supplied then the calling of this function is meaningless? i'm not sure what it means to return amount. the consumer should not use any values returned

bradleySuira: Hi, it was because if you send a null/undefined rate we got an error 
`return BN(amount).dividedBy(**rate**).dp(VALUE_DECIMALS, BN.ROUND_FLOOR)`

PR #56, opened 2021-02-17
Fix agent update

bradleySuira:


src/store/actions/performNextAction/swap.js
```diff
@@ -72,11 +72,33 @@ async function initiateSwap ({ getters, dispatch }, { order, network, walletId }
...
+    order.fromAddress,
+    order.secretHash,
+    order.swapExpiration,
+    order.fee
+  )
+
+  console.log('ORDER FUNDED', fundTx)
+
+  return {
+    status: 'INITIATION_REPORTED'
```
monokh, line 93: The tx obj also needs to be committed to state so that later it can be used to bump fee

bradleySuira: got it

bradleySuira: fixed

src/store/actions/performNextAction/swap.js
```diff
@@ -72,11 +72,32 @@ async function initiateSwap ({ getters, dispatch }, { order, network, walletId }
...
+    order.fromAmount,
+    order.fromCounterPartyAddress,
+    order.fromAddress,
+    order.secretHash,
+    order.swapExpiration,
+    order.fee
+  )
+
+  return {
+    fundTxHash: fundTx?.hash,
```
monokh, line 91: Should also commit the full object like the other transactions
for ethereum full object needs to be passed to `updateTransactionFee` to ensure it can do this

monokh: also should name this consistent to others 
`fromFundTx` already exists for intiiation
`fromSecondaryFundTx` is what you would call it to be consistent with how the agent calls it

src/views/Swap.vue
```diff
@@ -675,33 +675,39 @@ export default {
...
+        console.log('toTxTypes', toTxTypes)
+        const toFee = toTxTypes.reduce((accum, tx) => {
+          return accum.plus(getTxFee(this.toAsset, tx, toAssetFee))
+        }, BN(0))
+
+        fees[this.toAssetChain] = toFee
 
         if (this.sendTo) {
-          const sendFee = getTxFee(this.toAsset, TX_TYPES.SEND, feePrice)
+          const sendFee = getTxFee(this.toAsset, TX_TYPES.SEND, toAssetFee)
```
monokh, line 703: Is sendto fee being accounted for twice?
It looks like this would already be included because on `toFee` is also accumulates toTxTypes = `TX_TYPES.SEND`

bradleySuira: got it, let me fix it

PR #65, opened 2021-03-15
Seed phrase authentication feature

carlsachs:
EDIT 3/16/21: Ready for re-review

Absolute positioning taken out where it had been unless absolutely needed (eye logo on reveal page and checkbox/buttons on password page)


src/views/SeedPhrase/Warning.vue
```diff
@@ -0,0 +1,59 @@
...
+</script>
+
+<style lang="scss">
+
+.warning-phrase {
+  &_warning-line {
+    border: 3px solid $danger;
+  }
+  &_top-text {
+    color: $color-text-primary;
```
monokh, line 47: This seems to be the default color

src/views/SeedPhrase/Warning.vue
```diff
@@ -0,0 +1,59 @@
...
+    text-align: center;
+  }
+  &_eye-svg {
+    height: 150px;
+    width: 150px;
+  }
+  &_bottom-text {
+    text-align: center;
```
monokh, line 55: you can avoid creating a class for this and just add the bootstrap class `text-center` to achieve this 🙂

src/views/SeedPhrase/Warning.vue
monokh: if this is removed, you can simply add `text-center` and you won't need a class

PR #104, opened 2021-05-07
Search asset fix

carlsachs:
Fixed the search, you just have to press enter, maybe take a look at my @change in the input, maybe something wrong there?


carlsachs:
I think it was the ' || ' statement, but if not, then it worked the whole time and none of us pressed enter.

carlsachs:
Overview:

1) Fixes 0x BTC address error
 - Changed the address function to a different logic
2) Fixes the 'Max' amount not being changed in a swap on fee change
 - Added the else-if option to the logic, which changed and fixed the error
3) Fixes Liquidity Message
 - Will not show unless there is no market now.  The logic before was showing the message even when we had liquidity.
 - Talked to Thessy and she said to include it when we don't have the liquidity, or don't trade it, so that would be the !market, no?
4) Fixes Token Search in Manage Assets
 - Created logic to handle the sorting of assets onKeyup instead of on 'enter' press
  - Created a 'no token' page, which is not functional in it's links yet, have to ask Thessy what she wants there, but it looks good.

src/build.config.js
```diff
@@ -24,8 +24,8 @@ export default {
     ]
   },
   agentEndpoints: {
-    testnet: ['https://liquality.io/swap-testnet/agent'],
-    mainnet: ['https://liquality.io/swap/agent']
+    testnet: ['https://liquality.io/swap-testnet-dev/agent'],
```
monokh, line 27: shouldn't be checked in, it will update prod

src/views/ManageAssets.vue
```diff
@@ -5,11 +5,25 @@
...
+          <h4 class="manage-assets_noneText">Can't find this token</h4>
+          <h4 class="manage-assets_customText">Add Custom ERC20 tokens.</h4>
+          <h5>Learn how</h5>
+        </div>
+      </div>
     <div class="manage-assets_list">
+       <div v-if="assets.length === 0" class="mt-5 d-flex">
+        <div class="manage-assets_noneBottom">
+          <h5 class="manage-assets_noneBottomText">INQUIRE</h5>
+          <h4 class="manage-assets_bottomEnable">Enable other tokens</h4>
```
monokh, line 22: Where are the links?
These shouldn't have custom classes by the way, it should just use `<a>` tag
a link should be a link. If design presents the link differently, it's being inconsistent

src/views/ManageAssets.vue
```diff
@@ -75,11 +89,9 @@ export default {
         this.assets = assets
       } else {
         this.assets = assets.filter(
-          asset => asset.toLowerCase().includes(
-            this.search.toLowerCase()
-          ) ||
-        cryptoassets[asset]?.name.toLowerCase()
-        .includes(this.search.toLowerCase())
+          asset => asset.toUpperCase().includes(
```
monokh, line 92: What happened here? Why don't we match on name anymore ?

src/views/ManageAssets.vue
```diff
@@ -105,6 +117,35 @@ export default {
   flex-direction: column;
   min-height: 0;
 
+  &_noneTop {
```
monokh, line 120: shouldn't need this, use <a> tag

src/views/ManageAssets.vue
```diff
@@ -105,6 +117,35 @@ export default {
...
+
+  &_noneText {
+    font-weight: bold;
+  }
+
+  &_customText {
+    font-weight: 100;
+  }
+
+  &_noneBottom {
```
monokh, line 137: shouldn't use fixed height
we would need to adjust this value anytime we change the content - not good

This container should always be flex such that the bottom sticks

src/views/ManageAssets.vue
```diff
@@ -105,6 +117,35 @@ export default {
...
+    height: 190px;
+  }
+
+  &_bottomEnable, &_bottomLiquidity {
```
monokh, line 141: again doesn't seem required. use link

src/views/Swap/Swap.vue
```diff
@@ -429,7 +429,7 @@ export default {
       return this.$route.query.source || null
     },
     showNoLiquidityMessage () {
-      return !this.market || BN(this.min).gt(this.max)
```
monokh, line 432: why was the last part removed? it makes sure no liquidity message is displayed if there is not enough for a minimum swap

src/views/Swap/Swap.vue
```diff
@@ -880,6 +880,8 @@ export default {
       handler (val) {
         if (this.amountOption === 'max') {
           this.sendAmount = this.max
+        } else if (this.amountOption === 'min') {
```
monokh, line 883: does this fix something? Doesn't seem to make a different

monokh (changes requested):
The padding on the list of assets has been messed up and the list is not scrollable anymore
image

src/views/Swap/Swap.vue
monokh: If it's fixing the max issue, I can still reproduce it after this change

src/views/ManageAssets.vue
carlsachs: Yeah these will be changed, need to talk to thessy about them

PR #121, opened 2021-06-14
Pixel perfect

carlsachs:
Fixes to UI:

-Confetti changed
-onboarding flow has been redone per Thessy request
-Other small changes from the figma

### DIFF
--- a/src/views/Onboarding/OnboardingHome.vue
+++ b/src/views/Onboarding/OnboardingHome.vue
@@ -536,8 +536,9 @@
           <span class="c20"></span>
         </p>
       </div>
-      <div class="onboading-home_tnc__actions">
-        <button class="btn btn-primary btn-lg btn-block mb-2" @click="acceptTnC">I Accept</button>
+      <div class="button-group">
+          <router-link :to="`/`"><button class="btn btn-light btn-outline-primary btn-lg">Cancel</button></router-link>
+          <button class="btn btn-primary btn-lg px-2" @click="acceptTnC">I Accept</button>
       </div>
     </div>
   </div>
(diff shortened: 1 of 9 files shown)


src/views/Onboarding/OnboardingSetup.vue
```diff
@@ -84,6 +88,19 @@ export default {
   padding: 0 !important;
   overflow: hidden;
 
+  &_description {
+    font-weight: 500;
+    font-size: 14px;
+  }
+
+  &_logo-wrap {
```
monokh, line 96: Relative and absolute position should be avoided unless 100% necessary
In this case, you can adjust the margin of this element or elements below it to achieve what you need

src/views/Onboarding/OnboardingSetup.vue
```diff
@@ -84,6 +88,19 @@ export default {
   padding: 0 !important;
   overflow: hidden;
 
+  &_description {
+    font-weight: 500;
```
monokh, line 92: font-weight 500 is not doing anything on screen for me

there are only 2 weights, normal (not defined) and bold (700)

src/views/Onboarding/OnboardingPassword.vue
```diff
@@ -93,6 +93,13 @@ export default {
     margin-top: 2rem;
   }
 
+  &_errorLength {
+    background: #fff;
+    border-top: 1.75px solid #F12274;
```
monokh, line 98: `1.75px` should be either 1 or 2
any color used should be a sass variable. for errors there is `$danger`

src/views/Onboarding/OnboardingHome.vue
```diff
@@ -567,6 +568,13 @@ export default {
 .onboading-home {
   padding: 70px 0 0 0 !important;
 
+  &_buttonGroup {
```
monokh, line 571: This isn't used on screen at all?

monokh (changes requested):

monokh (approved):

PR #127, opened 2021-06-18
feat: Added some element ID's and added simple mocha spec

beemi:
### Improvements

- Added element ID's in different component pages for writing UI Tests
- Added Mocha test runner added

### Test

- Added tests/utils/TestUtils.js responsible for generating root URL
- `tests/integration/simple.spec.js` had 3 mocha tests 

1. Create new wallet
2. Import wallet with 12 words
3. Import wallet which has coins, need to pass `SEED_WORDS` as env


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname $0)/_/husky.sh"
 
-npm run lint
\ No newline at end of file
+#npm run lint
```
monokh, line 4: can we have this back? 😄

src/views/Account.vue
```diff
@@ -32,13 +32,13 @@
           </button>
         </div>
         <div class="account-container_actions">
-          <router-link :to="`/accounts/${accountId}/${asset}/send`"><button class="account-container_actions_button">
+          <router-link :to="`/accounts/${accountId}/${asset}/send`"><button class="account-container_actions_button" id="send">
             <div class="account-container_actions_button_wrapper"><SendIcon class="account-container_actions_button_icon" /></div>Send
           </button></router-link>
-          <router-link :to="`/accounts/${accountId}/${asset}/swap`"><button class="account-container_actions_button">
+          <router-link :to="`/accounts/${accountId}/${asset}/swap`"><button class="account-container_actions_button" id="swap">
```
monokh, line 38: Should things like "swap" should be more specific incase it conflicts with other IDs in the app?

monokh (approved):

src/views/Account.vue
beemi: This should work on the wallet asset screen, but the idea to add an assert name to it so that it will be unique, will do that in next

PR #136, opened 2021-06-21
[TEST]-Added BTC send to invalid address and check Review button has been di…

beemi:
### Improvements 💶

- Included page object pattern in the Puppeter framework
- Each page class should have page actions

### ### TEST 1 🎯
- Added Element ID for SEND view
- BTC send to an invalid address and check the error message
- Check Review button has been disabled
### ### TEST 2 🎯
- Try to SEND with exceed amount and check the error message
- Send Review button has been Disabled
### docs 📚
- Updated Readme docs with tests
- Fix some existing docs
### dependency upgrade 🛠

- Upgrade eslint to latest version


tests/integration/testnet/testNetNetwork.spec.js
```diff
@@ -532,4 +531,207 @@ describe('Liquality wallet...', async () => {
...
+    const walletStatText = await page.$eval('.wallet-stats', el => el.innerText)
+    expect(walletStatText, 'Wallet stats has currency should be USD').contain('USD')
+
+    // Check the Total amount
+    await page.waitForTimeout(10000)
+    const totalAmount = await page.$eval('.wallet-stats_total', el => (el.innerText).replace(/[.,\s]/g, ''))
+    expect(parseInt(totalAmount), 'Funds in my wallet should be greater than 2000 USD').greaterThanOrEqual(2000)
+    console.log(chalk.green('After Import wallet, the funds total greater than 2000 USD'))
+
+    // Click on SEND Option
```
monokh, line 609: Everything above this seems to be outwith the scope of the current test,
If there's something that needs to be done before each test, it should be abstracted out into a utility to avoid too much duplication. In that way, if we update something to do with importing/password, we just need to update once. 
Do you agree with this pattern?

monokh (changes requested):

tests/integration/testnet/testNetNetwork.spec.js
beemi: Yes, @monokh  I have a separate task to put these tests in Page object pattern so that we will keep page actions in each Class then re use them across the test suite.

beemi: Just for now Trying to add some coverage for the use cases, then i will break down these into each page object util

bradleySuira (approved):
LGTM

PR #137, opened 2021-06-22
[TEST]-Added SEND test case from SOV to ETH

beemi:
- Generate random ETH address
- Send SOV to ETH address
- Validate the Transaction details Page


bradleySuira (approved):
LGTM

PR #138, opened 2021-06-23
[TEST] - SEND feature

beemi:
### TESTS

- SEND test cases
- SEND tests with invalid format address * exceeded limit error message checks

### Pages

- Added Page in tests  for each view

### Test utils

- Generate random ETH and BTC address for tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "@vue/eslint-config-standard": "^5.1.2",
     "@vue/test-utils": "^1.1.0",
     "babel-eslint": "^10.1.0",
+    "bitcoin-address-generator": "^1.0.2",
     "eslint": "^7.29.0",
     "eslint-plugin-import": "^2.23.4",
     "eslint-plugin-node": "^11.1.0",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -89,6 +89,7 @@
     "@vue/eslint-config-standard": "^5.1.2",
     "@vue/test-utils": "^1.1.0",
     "babel-eslint": "^10.1.0",
+    "bitcoin-address-generator": "^1.0.2",
```
monokh, line 92: Can we avoid adding a new module for this one? From a security perspective, it's best to keep dependencies as low as possible

monokh: I think it should be possible to use just `bitcoinjs-lib` which is already included in the wallet

monokh: I found an example: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/integration/addresses.spec.ts#L13

monokh (changes requested):

package.json
beemi: Removed that dependencies and used existing one

monokh (approved):

PR #139, opened 2021-06-23
Fix BTC to BTC SEND test case

beemi:
- BTC Generate testnet Random address

- SEND BTC wallet to another BTC wallet


bradleySuira (approved):
LGTM

PR #151, opened 2021-06-29
Arbitrum integration

tima-t:
Integration of arbitrum in wallet 
- added update for coingeckoService when coingeckoId is the same for a few native assets


src/store/utils.js
```diff
@@ -51,6 +52,9 @@ export async function getPrices (baseCurrencies, toCurrency) {
   const { data } = await axios.get(`${COIN_GECKO_API}/simple/price?ids=${coindIds.join(',')}&vs_currencies=${toCurrency}`)
   let prices = mapKeys(data, (v, coinGeckoId) => findKey(cryptoassets, asset => asset.coinGeckoId === coinGeckoId))
   prices = mapValues(prices, rates => mapKeys(rates, (v, k) => k.toUpperCase()))
+  for (const code in nativeCodeSynonyms) {
```
monokh, line 55: I think this would work the same if you use `baseCurrencies` to do the lookup on the prices returned from coingecko and then nativeCodeSynonyms would not be required

src/utils/fees.js
```diff
@@ -10,7 +10,8 @@ const SEND_FEE_UNITS = {
   BNB: 21000,
   NEAR: 10000000000000,
   MATIC: 21000,
-  ERC20: 90000
+  ERC20: 90000,
+  ARBETH: 2000000
```
monokh, line 14: `2000000` gas for sending in arbitrum? Is this correct?
Seems to be the same as initiating a swap - doesn't seem right

monokh (changes requested):

src/utils/fees.js
tima-t: That was suggestion from the tech team to use slightly higher gas limit, because the gas price could vary and I left it for safety reasons. 

But from my observations it is taking around   609500 Gas, so I can put it to 700 000

src/store/utils.js
tima-t: I will try it, It should work I guess

monokh (approved):
Looks good. Can you just make sure tests are passing?

monokh:
I think we need to update the `ethereum-networks` version

src/swaps/liquality/LiqualitySwapProvider.js
```diff
@@ -504,7 +505,8 @@ class LiqualitySwapProvider extends SwapProvider {
       BNB: 45000,
       MATIC: 45000,
       NEAR: 8000000000000,
-      ERC20: 100000
+      ERC20: 100000,
+      ARBETH: 2000000
```
monokh, line 509: Is this correct? It doesn't seem right that the claiming takes the same amount as gas as for initiating

PR #155, opened 2021-06-29
[Test]- Mocha grep Mainnet tests

beemi:
- Run all tests against PR creation
- Run selected tests against build release tag
- Added moch grep


monokh (approved):
although test failed

PR #170, opened 2021-07-08
[TEST]-Added receive tests

beemi:


tests/integration/receive.spec.js
```diff
@@ -164,4 +195,102 @@ describe('Liquality wallet- Receive-["mainnet"]', async () => {
     // After done
     await overviewPage.CheckAssertOverviewDetails(page, nearPlatform)
   })
+  it('Check Receive for ARBETH', async () => {
```
monokh, line 198: Looks like a lot of duplication here. Is it necessary? Could it use a generic function where you pass the asset and it will return the test?

monokh (approved):

tests/integration/receive.spec.js
beemi: Yes was thinking about that, will refactor that

beemi: Remove those duplicated code and added generic function

monokh (approved):

PR #175, opened 2021-07-12
Integration 1inch

tima-t:


src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    return {
+      id: uuidv4(),
+      fee: quote.fee,
+      slippage: 50,
+      ...updates
+    }
+  }
+
+  async estimateFees ({ network, walletId, asset, accountId, txType, quote, feePrices, max }) {
+    if (txType in OneinchSwapProvider.feeUnits) {
```
monokh, line 131: I found that one inch also returns "estimateGas" in the response for a quote
Can we use this to determine the fee estimate?

We'll need to add a buffer, (maybe * 1.5 ) 
because the gas estimate when actually sending the transaction happens inside the CAL and it is also multiplied by 1.5

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    if (!isEthereumChain(from) || !isEthereumChain(to) || amount <= 0) return null
+    const fromAmountInUnit = BN(currencyToUnit(cryptoassets[from], BN(amount)))
+    const chainIdFrom = ChainNetworks[cryptoassets[from].chain][network].chainId
+    const chainIdTo = ChainNetworks[cryptoassets[to].chain][network].chainId
+    if (chainIdFrom !== chainIdTo || !chainToRpcProviders[chainIdFrom]) return null
+
+    const trade = await axios({
+      url: this.agent + `/${chainIdFrom}/quote`,
+      method: 'get',
+      params: { fromTokenAddress: assets[from].contractAddress || nativeAssetAddress, toTokenAddress: assets[to].contractAddress || nativeAssetAddress, amount: fromAmountInUnit.toNumber() }
```
monokh, line 44: We should add a slippage here, static at 50 bips for now

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+
+    const callData = await axios({
+      url: this.agent + `/${chainId}/approve/calldata`,
+      method: 'get',
+      params: { tokenAddress: cryptoassets[quote.from].contractAddress }
+    })
+
+    const account = this.getAccount(quote.fromAccountId)
+    const client = this.getClient(network, walletId, quote.from, account?.type)
+    const approveTx = await client.chain.sendTransaction(callData.data)
```
monokh, line 82: Some of the parameters from `callData.data` cannot go into the signature for `sendTransaction`
I think it's better to set it explicitly.

Also need to add the fee properly here. in `callData` it's `gasPrice` but the correct parameter is `fee` and it needs to be in GWEI

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    const fromAddressRaw = await this.getSwapAddress(network, walletId, quote.from, quote.fromAccountId)
+    const fromAddress = chains[toChain].formatAddress(fromAddressRaw)
+
+    const trade = await axios({
+      url: this.agent + `/${chainId}/swap`,
+      method: 'get',
+      params: { fromTokenAddress: assets[quote.from].contractAddress || nativeAssetAddress, toTokenAddress: assets[quote.to].contractAddress || nativeAssetAddress, amount: quote.fromAmount, fromAddress: fromAddress, slippage: 0.5 }
+    })
+    await this.sendLedgerNotification(quote, account, 'Signing required to complete the swap.')
+    const swapTx = await client.chain.sendTransaction(trade.data.tx)
```
monokh, line 108: Same as above about being explicit and adding fee

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    if (allowance.gte(inputAmount)) {
+      return {
+        status: 'APPROVE_CONFIRMED'
+      }
+    }
+
+    const callData = await axios({
+      url: this.agent + `/${chainId}/approve/calldata`,
+      method: 'get',
+      params: { tokenAddress: cryptoassets[quote.from].contractAddress }
```
monokh, line 77: This is doing an infinite approval I think? You should specify the quote amount `quote.fromAmount` here.

src/swaps/oneinch/OneinchSwapProvider.js
monokh: The fee (gas Price in gwei) selected by the user is under `quote.fee`

src/swaps/oneinch/OneinchSwapProvider.js
monokh: I see you are adding a slippage on swap 👍

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    const client = this.getClient(network, walletId, quote.from, account?.type)
+    const fromAddressRaw = await this.getSwapAddress(network, walletId, quote.from, quote.fromAccountId)
+    const fromAddress = chains[toChain].formatAddress(fromAddressRaw)
+
+    const trade = await axios({
+      url: this.agent + `/${chainId}/swap`,
+      method: 'get',
+      params: { fromTokenAddress: assets[quote.from].contractAddress || nativeAssetAddress, toTokenAddress: assets[quote.to].contractAddress || nativeAssetAddress, amount: quote.fromAmount, fromAddress: fromAddress, slippage: 0.5 }
+    })
+    await this.sendLedgerNotification(quote, account, 'Signing required to complete the swap.')
```
monokh, line 107: The quote displayed to the user from `/quote` call is different than the price executed later using `/swap`
So we should add an extra check here to compare `trade.toTokenAmount` to `quote.toAmount`. If it's larger than the slippage, throw an error

src/swaps/oneinch/OneinchSwapProvider.js
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    return updates
+  }
+
+  static txTypes = {
+    SWAP: 'SWAP'
+  }
+
+  static feeUnits = {
```
monokh, line 200: this should be removed after fee estimate comes from quote

monokh (changes requested):

src/swaps/oneinch/OneinchSwapProvider.js
tima-t: Yes, this is a good idea! Will add it

src/swaps/oneinch/OneinchSwapProvider.js
tima-t: Yeah, I thought that it will be more user friendly to use infinite approve, rather than approving some small amount every time, but I can update it if that is the requirement

src/swaps/oneinch/OneinchSwapProvider.js
tima-t: Makes sense

src/swaps/oneinch/OneinchSwapProvider.js
tima-t: will remove it

src/swaps/oneinch/OneinchSwapProvider.js
monokh: It's a fair point. I can see the advantages on both sides

monokh (approved):

PR #177, opened 2021-07-13
Trello fixes

carlsachs:
Pull fixes three bugs:

Uneven receive faucet links:

https://trello.com/c/AdB7ypJi/210-dev-testnet-alignment-issue
https://trello.com/c/ORPhGvoS/208-dev-ux-issue-on-arbeth-receiver-screen
https://trello.com/c/bTLAeLy0/204-arbitrum-test-wallet-link-is-not-visible-properly

Implements a warning for high transaction fees (Could NOT TEST due to not having mainnet funds in the wallet, and no testnet liquidity, but the logic says it should check out, will send to QA to test after Mo confirms logic):

https://trello.com/c/L1JuThSK/128-user-needs-to-be-warned-of-high-transaction-fees

Makes sure no errors are shown on an input when the "No liquidity" message is present:

https://trello.com/c/AJTG4Md3/163-prod-min-n-max-message-still-present-when-there-is-not-liquidity-for-eth-to-btc


src/views/Swap/Swap.vue
```diff
@@ -765,6 +770,11 @@ export default {
       delete this.customFees[asset]
       this.resetFees()
     },
+    highFeesCheck (highFees, fromSwapFee, sendAmount) {
```
monokh, line 773: Where is this used?

src/views/Swap/Swap.vue
```diff
@@ -566,6 +567,10 @@ export default {
     amountError () {
       const amount = BN(this.safeAmount)
 
+      if (this.showNoLiquidityMessage === true) {
```
monokh, line 570: better written `if (this.showNoLiquidityMessage)`

monokh: also should be at the top of the function since nothing needs to happen if it returns

monokh (changes requested):

src/views/Swap/Swap.vue
```diff
@@ -429,6 +429,9 @@ export default {
     account () {
       return this.accountItem(this.fromAccountId)
     },
+    isHighFee () {
+      return this.toSwapFee >= this.sendAmount * 0.25
```
monokh, line 433: why only `toSwapFee`? That is the fee paid only for the receiving side of the swap. `fromSwapFee` is another fee which will be paid as part of the swap.

you also can't compare sendAmount to `toSwapFee` because they will be amounts for 2 different assets.
for example 
1 BTC - 50 ETH

`toSwapFee` will be the fee paid to claim ETH
`sendAmount` is 1 BTC
so this comparison cannot be made

I think you need to compare something like the total fee in fiat vs the receive amount in fiat

monokh (changes requested):

carlsachs:
Changed up the logic on the negative swap/high fees (tested in testnet but couldn't test negative because no mainnet, although I'm sure it works just fine), fixed small UI error/s thessy pointed out, fixed the package lock issue with the 36000 additions/deletions.  Sorry for late push, was waiting on the UI for the error messages , didn't get that til Thursday/Friday @monokh

src/views/Swap/Swap.vue
```diff
@@ -651,6 +668,14 @@ export default {
         [this.assetChain]: this.asset,
         [this.toAssetChain]: this.toAsset
       }
+    },
+    checkHighFee () {
```
monokh, line 672: since it will become a property, it should not be named with an action.
more like "isFeeTooHigh"
same with other one "isSwapNegative"

src/build.config.js
```diff
@@ -46,7 +46,7 @@ export default {
         name: 'Liquality',
         icon: 'liquality.svg',
         type: SwapProviderType.LIQUALITY,
-        agent: process.env.VUE_APP_AGENT_TESTNET_URL || 'https://liquality.io/swap-testnet/agent'
+        agent: process.env.VUE_APP_AGENT_TESTNET_URL || 'https://liquality.io/swap-testnet-dev/agent'
```
monokh, line 49: Should not be checked in

monokh (commented):
CI failing on lock file

harshjv:
@carlsachs can you run `npm install` and then commit updated `package-lock.json` file?

PR #179, opened 2021-07-13
Feature multiple account selection

bradleySuira:


src/broker/Background.js
```diff
@@ -152,8 +152,16 @@ class Background {
...
-    const allowed = this.externalConnectionApprovalMap[origin]
+    let chain
+    if (data.chain) {
+      chain = data.chain
+    } else {
+      const { asset } = data
+      chain = assets[asset].chain
+    }
+    const { externalConnections, activeWalletId } = this.store.state
+    const allowed = Object.keys(externalConnections[activeWalletId] || {}).includes(origin) &&
```
monokh, line 163: Shouldn't it be `|| []` ? I think object does not have includes

src/store/actions/hardware-wallet/getLedgerAccounts.js
```diff
@@ -5,7 +5,16 @@ export const getLedgerAccounts = async (
   { network, walletId, asset, walletType, startingIndex, numAccounts }
 ) => {
   const { client, networkAccounts } = getters
-  const _client = client(network, walletId, asset, walletType, 0, false)
+  const _client = client(
+    {
+      network,
+      walletId,
+      asset,
+      walletType,
```
monokh, line 13: Hmm, isn't the getter now take `accountId` instead of type and index?

src/store/actions/requestPermission.js
```diff
@@ -27,6 +27,7 @@ const ALLOWED = [
 ]
 
 export const requestPermission = async ({ state, dispatch }, { origin, data }) => {
+  await dispatch('requestUnlockWallet')
```
monokh, line 30: I think i had this at some point then removed it. can you just double check if it causes issues by trying some dapps when your wallet is lockeD?

monokh (changes requested):
Pretty much there, can you also update the `OneInchSwapProvider`?

src/broker/Background.js
bradleySuira: got it I will make the same inside OneInchSwapProvider, about the includes it's because Im using Object.keys that returns an array

bradleySuira: @monokh I can't found the `OneInchSwapProvider ` in this branch also in dev

src/store/actions/requestPermission.js
bradleySuira: Hi checked, but requestUnlockWallet only check. if the wallet is locket if not just return and do anything

src/store/actions/hardware-wallet/getLedgerAccounts.js
bradleySuira: Hi good point, but you can still send the wallet type if you don't send the accountId, BTW I removed this call because we don't need it
input parameters:
```
{
      network,
      walletId,
      asset,
      accountId,
      useCache = true,
      walletType = 'default',
      index = 0
    }
```

then we check for the default values 
```
const accountType = account?.type || walletType
const accountIndex = account?.index || index
```

PR #186, opened 2021-07-26
Feat/fastbtc integration

tima-t:


src/build.config.js
```diff
@@ -80,6 +80,11 @@ export default {
           bsc: '0x3a712CC47aeb0F20A7C9dE157c05d74B11F172f5'
         },
         referrerFee: 0.3
+      },
+      fastBTC: {
+        name: 'FastBTC',
+        type: SwapProviderType.FASTBTC,
+        agent: 'http://3.131.33.161:3000/'
```
monokh, line 87: `agent` is specific to liquality
should be sometihng more appropriate like `bridgeEndpoint`

src/swaps/fastbtc/FastbtcSwapProvider.js
```diff
@@ -0,0 +1,254 @@
...
+      console.log('FatBtc socket connected')
+    })
+
+    this.socketConnection.on('disconnect', function () {
+      console.log('FastBtc socket disconnected')
+    })
+  }
+
+  async getSupportedPairs () {
+    return []
```
monokh, line 32: Follow what `LiqualitySwapProvider` does for this one
`[]` should be returned when markets are unbounded. E.g. in AMMs like uniswap. In this case it's only BTC -> RBTC

src/swaps/fastbtc/FastbtcSwapProvider.js
```diff
@@ -0,0 +1,254 @@
...
+        resolve(res)
+      })
+    })
+  }
+
+  async getQuote ({ network, from, to, amount }) {
+    const fromAmountInUnit = BN(currencyToUnit(cryptoassets[from], BN(amount)))
+    const validAmountRange = await this._getTxAmount()
+    const isQuoteAmountInTheRange = amount <= validAmountRange.max && amount >= validAmountRange.min
+    if (from !== 'BTC' || to !== 'RBTC' || !isQuoteAmountInTheRange) return null
```
monokh, line 69: from and to can be checked at the beginning to avoid any other calls if not needed

src/swaps/fastbtc/FastbtcSwapProvider.js
```diff
@@ -0,0 +1,254 @@
...
+      fromAmount: fromAmountInUnit.toNumber(),
+      toAmount: toAmountInUnit.toNumber()
+    }
+  }
+
+  async sendSwap ({ network, walletId, quote }) {
+    if (quote.from !== 'BTC' || quote.to !== 'RBTC') return null
+    const toChain = cryptoassets[quote.to].chain
+    const account = this.getAccount(quote.fromAccountId)
+    const client = this.getClient(network, walletId, quote.from, account?.type)
```
monokh, line 86: I don't think this is like this anymore? I think that you need to pass the accountid not type. Try to get latest from dev to see

src/swaps/fastbtc/FastbtcSwapProvider.js
```diff
@@ -0,0 +1,254 @@
...
+    return updates
+  }
+
+  static txTypes = {
+    SWAP_INITIATION: 'SWAP_INITIATION'
+  }
+
+  static feeUnits = {
+    SWAP_INITIATION: {
+      BTC: 290,
```
monokh, line 199: is feeUnits even required here at all? BTC is calculated dynamically, and you can only go BTC -> RBTC and you don't need to pay fees for RBTC. So only estimation for BTC is required

src/swaps/fastbtc/FastbtcSwapProvider.js
```diff
@@ -0,0 +1,254 @@
...
+      case 'WAITING_FOR_RECEIVE':
+        updates = await withInterval(async () => this.waitForReceive({ swap, network, walletId }))
+        break
+    }
+
+    return updates
+  }
+
+  static txTypes = {
+    SWAP_INITIATION: 'SWAP_INITIATION'
```
monokh, line 194: should be called just `SWAP` I guess. `iniitation` implies there are more steps

monokh (changes requested):

monokh (approved):

PR #190, opened 2021-07-29
feat: Fetch token details

AndonMitev:
feat: Fetch token details


kraikov (commented):
LGTM 👍

src/utils/asset.js
```diff
@@ -135,6 +142,131 @@ export const getExplorerTransactionHash = (asset, hash) => {
...
+        postData(this.sourceUrl(), this.rpcBody(contractAddress).decimals)
+      ])
+
+      return parseAssetData(_symbol.result, _name.result, _decimals.result)
+    }
+  }
+}
+
+const fetchData = async url => {
+  const { data } = await axios.get(url)
```
kraikov, line 249: What if the request fail?

src/utils/asset.js
```diff
@@ -135,6 +142,131 @@ export const getExplorerTransactionHash = (asset, hash) => {
...
+
+  return data
+}
+
+const postData = async (url, body) => {
+  const { data } = await axios.post(url, body)
```
kraikov, line 255: What if the request fail?

src/utils/asset.js
```diff
@@ -135,6 +142,131 @@ export const getExplorerTransactionHash = (asset, hash) => {
...
+        fetchData(this.sourceUrl(contractAddress).decimals)
+      ])
+      return parseAssetData(_symbol.result, _name.result, _decimals.result)
+    }
+  },
+  rsk: {
+    sourceUrl() {
+      return 'https://public-node.rsk.co'
+    },
+    rpcBody(contractAddress) {
```
monokh, line 196: I'm wondering, if we are doing like this for some of them, Why don't we just use this method for everything so there is one single implementation? We can keep it very tidy by using ethers too.

What is the benefit of using the apis?

src/views/CustomToken.vue
```diff
@@ -164,25 +164,25 @@ export default {
 
       if (this.existingAsset) {
         customToken = this.existingAsset
-      } else if (this.activeNetwork === 'mainnet') {
+      } else if (this.activeNetwork === 'mainnet' && this.contractAddress) {
         try {
-          const result = await axios.get(`https://api.ethplorer.io/getTokenInfo/${this.contractAddress}?apiKey=freekey`)
-          const { symbol, name, decimals } = result.data
-          customToken = { symbol, name, decimals: parseInt(decimals), chain: 'ethereum' }
+          const { symbol, name, decimals } = await TOKEN_DETAILS[this.chain].getDetails(this.contractAddress)
```
monokh, line 169: `TOKEN_DETAILS` makes it feel like this is some sort of constant value instead of a module 
i think it should be more like `tokenDetailProviders` or something like that

src/utils/abi.json
```diff
@@ -0,0 +1,41 @@
+[
```
monokh, line 1: name should be like tokenAbi?

src/utils/asset.js
```diff
@@ -135,6 +144,56 @@ export const getExplorerTransactionHash = (asset, hash) => {
...
+      return await fetchTokenDetails(contractAddress, 'https://rpc-mainnet.matic.network/');
+    }
+  },
+  rsk: {
+    async getDetails(contractAddress) {
+      return await fetchTokenDetails(contractAddress, 'https://public-node.rsk.co')
+    }
+  },
+  bsc: {
+    sourceUrl(contractAddress) {
```
monokh, line 169: What happened with binance? the same method doesn't work there?

monokh (approved):

PR #191, opened 2021-07-30
feat: Rsk legacy derivation path

AndonMitev:
…se legacy coin type other way apply the new one

### DIFF
--- a/src/store/actions/unlockWallet.js
+++ b/src/store/actions/unlockWallet.js
@@ -1,31 +1,10 @@
 import { encrypt, decrypt, decryptLegacy } from '../../utils/crypto'
-import { Client } from '@liquality/client'
-import { EthereumRpcProvider } from '@liquality/ethereum-rpc-provider'
+import { getLegacyRskBalance } from '../utils'
 
 export const unlockWallet = async ({ commit, state, dispatch }, { key }) => {
   let wallets = await decrypt(state.encryptedWallets, key, state.keySalt)
 
-  const accKeys = Object.keys(state.accounts);
-
-  let addresses = [];
-
-  for(let i = 0; i < accKeys.length; i++) {
-    const currentAccs = currentState.accounts[accKeys[i]].mainnet;
-    
-    for(let k = 0; k < currentAccs.length; k++) {
-      if(currentAccs[k].chain === 'rsk') {
-        addresses.push(...currentAccs[k].addresses)
-      }
-    }
-  }
-
-  const client = new Client()
-    .addProvider(
-      new EthereumRpcProvider({ uri: 'https://public-node.rsk.co' })
-    );
-  
-  const balance = await client._chain.getBalance(addresses)
-  
+  const balance = await getLegacyRskBalance(state.accounts)  
   // Migration to new encryption method
   // TODO: to be removed
   if (!wallets) {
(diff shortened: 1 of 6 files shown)


src/store/migrations.js
```diff
@@ -210,17 +210,38 @@ const migrations = [
...
+    migrate: async (state) => {
+      const accKeys = Object.keys(state.accounts);
+
+      if (!accKeys.length) {
+        return {
+          ...state,
+        }
+      }
+
+      const accData = state.accounts[accKeys[0]].mainnet;
```
monokh, line 225: `accounts` is indexed by wallet id. I think you need to iterate over the walletIds and override rsk for each of them. See other migrations how to do this

src/store/factory/client.js
```diff
@@ -93,7 +94,12 @@ function createEthereumClient (
   const ethClient = new Client()
   ethClient.addProvider(new EthereumRpcProvider({ uri: rpcApi }))
 
-  const derivationPath = `m/44'/${ethereumNetwork.coinType}'/${indexPath}'/0/0`
+  const legacyCoinType = 137;
+  const { rskLegacyDerivation } = state;
+  const coinType = rskLegacyDerivation ? legacyCoinType : ethereumNetwork.coinType
+
+  const derivationPath = `m/44'/${coinType}'/${indexPath}'/0/0`
```
monokh, line 101: Ledger wallet type should always use "legacy derivation"

monokh (changes requested):
I'm worried about the scenario where the user receives some balance and the wallet has not updated for some reason. In that case, the user's wallet will be moved to new deriv path but they actually have a balance in the old deriv path. Is it possible to fetch the actual balance instead of just from the state? 
 you can see `updateBalances` for how to get balance

AndonMitev:
> I'm worried about the scenario where the user receives some balance and the wallet has not updated for some reason. In that case, the user's wallet will be moved to new deriv path but they actually have a balance in the old deriv path. Is it possible to fetch the actual balance instead of just from the state?
> you can see `updateBalances` for how to get balance

All fixed

src/store/actions/unlockWallet.js
```diff
@@ -1,15 +1,38 @@
 import { encrypt, decrypt, decryptLegacy } from '../../utils/crypto'
+import { Client } from '@liquality/client'
+import { EthereumRpcProvider } from '@liquality/ethereum-rpc-provider'
 
 export const unlockWallet = async ({ commit, state, dispatch }, { key }) => {
   let wallets = await decrypt(state.encryptedWallets, key, state.keySalt)
 
+  const accKeys = Object.keys(state.accounts);
```
monokh, line 8: You can call it "walletIds" 😄

src/store/actions/unlockWallet.js
```diff
@@ -1,15 +1,38 @@
...
+
+  let addresses = [];
+
+  for(let i = 0; i < accKeys.length; i++) {
```
monokh, line 12: this can be more readable in a foreach, either `for(const walletId in walletIds)` or `walletIds.forEach(walletId => {})`
same for below loop

src/store/actions/unlockWallet.js
monokh: It might be nice to put this whole thing in a function like `getLegacyRskBalance(state.accounts)`

src/store/migrations.js
```diff
@@ -210,17 +212,53 @@ const migrations = [
...
+
+      if (!accKeys?.length) {
+        return {
+          ...state,
+        }
+      }
+
+      let addresses = [];
+
+      for(let i = 0; i < accKeys.length; i++) {
```
monokh, line 229: If you pull out the code in to a util function you can reuse it here

bradleySuira (approved):
LGTM

monokh (approved):

PR #202, opened 2021-08-05
Hotfix/wallet rsk path

AndonMitev:
## What?

#Notion/ Trello

## Testing?
- [ ] Run tests locally

## Screenshots (optional)
0

### DIFF
--- a/src/store/factory/client.js
+++ b/src/store/factory/client.js
@@ -39,6 +39,7 @@ import cryptoassets from '@/utils/cryptoassets'
 import buildConfig from '../../build.config'
 import { ChainNetworks } from '@/store/utils'
 import state from '../state'
+import store from '../../store'
 
 function createBtcClient (network, mnemonic, walletType, indexPath = 0) {
   const isTestnet = network === 'testnet'
--- a/src/store/factory/client.js
+++ b/src/store/factory/client.js
@@ -94,14 +94,15 @@ function createEthereumClient (
   const ethClient = new Client()
   ethClient.addProvider(new EthereumRpcProvider({ uri: rpcApi }))
 
-  const legacyCoinType = 137
-  const { rskLegacyDerivation } = state
-  let coinType
-
+  const legacyCoinType = '137'
+  const ethereumNetworkCoinType = '60'
+  const { rskLegacyDerivation } = store.state
+  let coinType = ethereumNetwork.coinType
+  
   if (walletType === 'rsk_ledger') {
     coinType = legacyCoinType
-  } else {
-    coinType = rskLegacyDerivation ? legacyCoinType : ethereumNetwork.coinType
+  } else if (ethereumNetwork.name === 'rsk_mainnet') {
+    coinType = rskLegacyDerivation || state.rskLegacyDerivation ? legacyCoinType : ethereumNetworkCoinType
   }
 
   const derivationPath = `m/44'/${coinType}'/${indexPath}'/0/0`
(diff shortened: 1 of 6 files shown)


src/store/utils.js
```diff
@@ -66,8 +67,21 @@ export const getLegacyRskBalance = async (accounts) => {
     .addProvider(
       new EthereumRpcProvider({ uri: 'https://public-node.rsk.co' })
     )
+    
+  if(mnemonic) {
```
monokh, line 71: At this point, why aren't we using `createClient` function? There's too much duplication here

src/store/utils.js
```diff
@@ -66,8 +67,21 @@ export const getLegacyRskBalance = async (accounts) => {
...
+        derivationPath: `m/44'/137'/${indexPath}'/0/0`
+      }))
+
+    const _addresses = await client._wallet.getAddresses();
+
+    addresses.push(..._addresses.map(e => e.address))
+  }
 
-  return await client._chain.getBalance(addresses)
+  return await client._chain.getBalance(addresses);
```
monokh, line 84: _chain is private
`client.chain`

src/store/factory/client.js
```diff
@@ -39,6 +39,7 @@ import cryptoassets from '@/utils/cryptoassets'
 import buildConfig from '../../build.config'
 import { ChainNetworks } from '@/store/utils'
 import state from '../state'
+import store from '../../store'
```
monokh, line 42: store I think is the right thing to use. 
`../state` will just return default values i think. Shouldn't use this

src/store/factory/client.js
```diff
@@ -94,14 +95,15 @@ function createEthereumClient (
...
+  const ethereumNetworkCoinType = '60'
+  const { rskLegacyDerivation } = store.state
+  let coinType = ethereumNetwork.coinType
+  
   if (walletType === 'rsk_ledger') {
     coinType = legacyCoinType
-  } else {
-    coinType = rskLegacyDerivation ? legacyCoinType : ethereumNetwork.coinType
+  } else if (ethereumNetwork.name === 'rsk_mainnet') {
+    coinType = rskLegacyDerivation || state.rskLegacyDerivation ? legacyCoinType : ethereumNetworkCoinType
```
monokh, line 106: this is getting confusing
rskLegacyDerivation === state.rskLegacyDerivation
why are there instances of both of them? And if they are the same value, what does `a || a ?` ... mean? When `a` is null, why do you need ternary to check it again?

monokh (changes requested):

src/store/utils.js
AndonMitev: Because createClient is used to set the new or old derivation path and this one is used always to log the user with the old derivation path and check the balance and base on that to determine if createClient() should use new or old derivation path

src/store/factory/client.js
AndonMitev: Yeah i notice that state and store.state are 2 different states and i just wanted to make sure i'm not gonna miss any value on both of them but yes i will remove state.rskLegacy if store.state is the real state

src/store/utils.js
```diff
@@ -66,8 +67,21 @@ export const getLegacyRskBalance = async (accounts) => {
     .addProvider(
       new EthereumRpcProvider({ uri: 'https://public-node.rsk.co' })
     )
+    
+  if(mnemonic) {
+    client.addProvider(
+      new EthereumJsWalletProvider({
+        network: ChainNetworks.ethereum.mainnet,
```
monokh, line 74: Should be rsk mainnet?

monokh (approved):

PR #218, opened 2021-08-16
Added new migration to force update for rsk legacy derivation path

bradleySuira:
Added new migration to force update for rsk legacy derivation path

## What?

https://www.notion.so/30ca4d49f3844a0fabdf4141f8a8bccc?v=44514edd52d14b719a9dec59d60147dd&p=900f7d3ace7d4d338fdf7a85306c09d9

### DIFF
--- a/src/store/migrations.js
+++ b/src/store/migrations.js
@@ -7,7 +7,8 @@ import { shouldApplyRskLegacyDerivation } from './utils'
 
 // add a migration to reuse
 const rskLegacyDerivationMigration = async (state) => {
-  if (!Object.keys(state.accounts)?.length) {
+  const hasAccounts = Object.keys(state.accounts || {}).length > 0
+  if (!hasAccounts) {
     return {
       ...state
     }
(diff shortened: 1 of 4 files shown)


src/store/migrations.js
```diff
@@ -5,6 +5,22 @@ import { chains, assets as cryptoassets } from '@liquality/cryptoassets'
 import { v4 as uuidv4 } from 'uuid'
 import { shouldApplyRskLegacyDerivation } from './utils'
 
+// add a migration to reuse
+const rskLegacyDerivationMigration = async (state) => {
+  if (!Object.keys(state.accounts)?.length) {
```
harshjv, line 10: No need to use `?.length`. also, can you make it more descriptive with variable names?

harshjv (changes requested):

src/store/migrations.js
bradleySuira: oh ok, I just moved this from the old code, let me check now. thanks!

bradleySuira: done, fixed @harshjv 

BR!

harshjv (approved):
👍

PR #227, opened 2021-08-23
fix: Handle custom fees

AndonMitev:
## What?
Use custom fee as multiplier 

#Notion/ Trello
https://www.notion.so/Unable-to-set-custom-fees-on-uniswap-dapp-a9167540fd864fcb85537a3126a920f6

### DIFF
--- a/src/views/PermissionSend.vue
+++ b/src/views/PermissionSend.vue
@@ -286,7 +286,7 @@ export default {
     },
     feeInUsdValue () {
       const gas = BN(this.request.args[0].gas, 16)
-      let feePerGas;
+      let feePerGas
       
       if (this.selectedFee === 'custom') {
         feePerGas = this.customFee
--- a/src/views/PermissionSend.vue
+++ b/src/views/PermissionSend.vue
@@ -9,7 +9,7 @@
         <p class="confirm-value" :style="getAssetColorStyle(asset)">{{symbol}}</p>
       </div>
       <div class="form-group">
-        <label>Transaction fee {{feeInUsdValue}} USD</label>
+        <label>Transaction fee {{usdValue}} USD</label>
       </div>
       </div>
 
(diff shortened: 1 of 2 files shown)


src/views/PermissionSend.vue
```diff
@@ -286,8 +286,15 @@ export default {
     },
     feeInUsdValue () {
       const gas = BN(this.request.args[0].gas, 16)
-      const feePerGas = BN(this.fees[this.activeNetwork]?.[this.activeWalletId]?.[this.assetChain][this.selectedFee].fee).div(1e9)
```
bradleySuira, line 289: Hi, maybe we can add a default value if we can't get the .fee field inside this.fees

BR!

AndonMitev: Do you have any in mind, for example 21000?

bradleySuira: O no sorry for not add details, I mean when we want to access to `this.fees[this.activeNetwork]?.[this.activeWalletId]?.[this.assetChain][this.selectedFee].fee` maybe we can get null/undefined and then `BN` constructor can throw an error.

src/views/PermissionSend.vue
```diff
@@ -210,6 +212,21 @@ export default {
         }
       }
     },
+    async estimateGas() {
```
monokh, line 215: Can you ensure that this is functioning correctly for other injections?
This permission screen i believe is shared among others so i'm wondering about something ETH specific being introduced here.

For example try a bitcoin send

some simple docs on bitcoin injection :

https://liquality.io/documentation/web_extension/bitcoin/getting_started/

src/views/PermissionSend.vue
```diff
@@ -237,7 +254,22 @@ export default {
       } else {
         this.updateSendFees(this.amount)
       }
-    }, 800)
+    }, 800),
+    async feeInUsdValue () {
+      const gas = await this.estimateGas();
```
monokh, line 259: I think it's not good to call this again

Ideally this function should be a computed property and use the component data as much as possible
Just calculating the USD value of whatever the fee is

monokh (changes requested):

src/views/PermissionSend.vue
AndonMitev: Tested and verified, works well for BTC

src/views/PermissionSend.vue
AndonMitev: Did it in ideally scenario :)

monokh (approved):

PR #231, opened 2021-08-24
fix: rsk derivation path and what's new modal

bradleySuira:


beemi (approved):

src/store/actions/getUnusedAddresses.js
```diff
@@ -15,20 +15,28 @@ export const getUnusedAddresses = async ({ state, commit, getters }, { network,
...
-        commit('UPDATE_ACCOUNT_ADDRESSES',
-          {
-            network,
-            accountId,
-            walletId,
-            asset,
-            addresses
-          })
+      let updatedAddresses = []
+      if (account.chain === 'bitcoin') {
```
monokh, line 21: Might want to re use the same logic in the other action but it's fine for now

monokh (approved):

PR #233, opened 2021-08-24
fix: Remove padding 0

AndonMitev:


src/views/Accounts/Create.vue
```diff
@@ -122,7 +122,6 @@ export default {
 
 <style lang="scss">
 .wrapper {
-  padding-top: 0 !important;
```
monokh, line 125: This was probably a style to change the padding only on this create screen
So it's probably still required. 

The bug is that it is applying to all other pages.
So I think you should figure out how to scope it to this page

either by scoping the style ( i think this will work?) 
<style lang="scss" scoped>

or more strict css selector

`.name-of-this-page .wrapper`

monokh (changes requested):

monokh (approved):

PR #235, opened 2021-08-24
feat: Add rsk fish token

bradleySuira:


beemi (approved):

src/store/actions/updateEnabledAssets.js
```diff
@@ -0,0 +1,13 @@
+import buildConfig from '../../build.config'
+import { Networks } from '../utils'
+
+export const updateEnabledAssets = async ({ state, dispatch, getters }) => {
```
monokh, line 4: wouldn't this renable assets even if you disabled them in settings?

bradleySuira: removed, thanks!

PR #240, opened 2021-08-25
Revamped swap layout with new logic and error messages

carlsachs:
) Added logic to check multiple values and compare them against others in computed.  Such as checking networks, gas, etc.
) Added components to deal with the error messaging (was able to squeese multiple errors into one
) Implemented that logic to sequence errors based on the users needs

Currently tidying up a third error (weird use case).  Check the chains in swap.vue within noTokens(), and also could you check mainnet?


src/views/Swap/Swap.vue
```diff
@@ -519,6 +527,13 @@ export default {
     networkWalletBalances () {
       return this.account?.balances
     },
+    sameNetwork () {
+      if (this.assetChain === this.toAssetChain) {
```
monokh, line 531: can be shortened to ` return this.assetChain === this.toAssetChain`

src/views/Swap/Swap.vue
```diff
@@ -587,24 +598,25 @@ export default {
...
-      if (this.assetChain === 'ETH') {
-        return !this.account?.balances?.ETH || this.account?.balances?.ETH === 0
+    noTokens () {
+      if (this.networkWalletBalances[this.asset] < this.min) {
+        return true
       }
-
-      if (this.toAssetChain === 'ETH') {
-        return !this.toAccount?.balances?.ETH || this.toAccount?.balances?.ETH === 0
+      if (this.networkWalletBalances[this.toAsset] < this.min) {
```
monokh, line 605: what does the `toAsset` have to do with the minimum ?
minimum is always denominated in the from asset

src/views/Swap/Swap.vue
```diff
@@ -12,11 +12,13 @@
...
-        <EthRequiredMessage :account-id="account.id"/>
+      <InfoNotification v-if="noTokens && sameNetwork">
+        <OneTokenRequired :accountId="accountId" :asset="asset" />
       </InfoNotification>
-
-      <InfoNotification v-else-if="showNoLiquidityMessage">
+      <InfoNotification v-if="noTokens">
+        <TokenRequiredMessage :accountId="account.id" :asset="asset" :toAsset="toAsset" />
+      </InfoNotification>
+      <InfoNotification v-else-if="!noTokens && showNoLiquidityMessage">
```
monokh, line 21: at this `else-if` 
`noTokens` will always be false

monokh: so you don't need to check it

src/components/OneTokenRequired.vue
```diff
@@ -0,0 +1,43 @@
+<template>
```
monokh, line 1: I'm actually not understanding why there is 2 components for this?
They both seem to say the same thing as well with the difference being a "Buy gas" button

src/views/Swap/Swap.vue
```diff
@@ -12,11 +12,13 @@
       >
         Swap
       </NavBar>
-      <InfoNotification v-if="ethRequired">
-        <EthRequiredMessage :account-id="account.id"/>
+      <InfoNotification v-if="noTokens && sameNetwork">
```
monokh, line 15: Why does something special need to be done for `sameNetwork`? 
Can you actually detail what behaviour you are trying to achieve? 
To my understanding: 
- Show {asset} balance required message when user swapping from {asset} and they don't have balance
- Show ETH balance required message when user swapping to ETH and they don't have balance

PR #245, opened 2021-08-27
fix: Solana Links

AndonMitev:
## What?

#Notion/ Trello

## Testing?
- [ ] Run tests locally

## Screenshots (optional)
0

### DIFF
--- a/src/utils/asset.js
+++ b/src/utils/asset.js
@@ -70,8 +70,8 @@ const EXPLORERS = {
   },
   solana: {
     testnet: {
-      tx: 'https://explorer.solana.com/tx/{template}?cluster=devnet',
-      address: 'https://explorer.solana.com/address/{template}?cluster=devnet'
+      tx: 'https://explorer.solana.com/tx/{hash}?cluster=devnet',
+      address: 'https://explorer.solana.com/address/{hash}?cluster=devnet'
     },
     mainnet: {
       tx: 'https://explorer.solana.com/tx/',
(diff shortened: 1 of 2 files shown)


src/utils/asset.js
```diff
@@ -136,7 +136,9 @@ export const getTransactionExplorerLink = (hash, asset, network) => {
 
 export const getAddressExplorerLink = (address, asset, network) => {
   const chain = cryptoassets[asset].chain
-  return `${EXPLORERS[chain][network].address}${address}`
+  const link = `${EXPLORERS[chain][network].address}`
+
+  return link.includes('{address}') ? link.replace('{address}', address) : link + address
```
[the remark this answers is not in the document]
AndonMitev: applied

src/utils/asset.js
```diff
@@ -131,12 +131,14 @@ export const getTransactionExplorerLink = (hash, asset, network) => {
   const chain = cryptoassets[asset].chain
   const link = `${EXPLORERS[chain][network].tx}`
 
-  return link.includes('{hash}') ? link.replace('{hash}', transactionHash) : link + transactionHash
+  return link.includes('{template}') ? link.replace('{template}', transactionHash) : link + transactionHash
```
monokh, line 134: This is not what i meant 😅 

I mean that it's unnecessary to have 2 ways of adding the hash/address to the URL
The placeholder approach works for everything
so we can add the placeholder `{hash}` to every url in `EXPLORERS`
and then simply do `link.replace('{hash}', transactionHash)`

src/utils/asset.js
```diff
@@ -131,12 +131,14 @@ export const getTransactionExplorerLink = (hash, asset, network) => {
...
 }
 
 export const getAddressExplorerLink = (address, asset, network) => {
   const chain = cryptoassets[asset].chain
-  return `${EXPLORERS[chain][network].address}${address}`
+  const link = `${EXPLORERS[chain][network].address}`
+
+  return link.includes('{template}') ? link.replace('{template}', address) : link + address
```
monokh, line 141: similar for this one

monokh (changes requested):

src/utils/asset.js
AndonMitev: applied

monokh (approved):

PR #249, opened 2021-08-30
fix: rsk derivation path

bradleySuira:
https://www.notion.so/Both-legacy-and-new-RSK-accounts-should-be-visible-e52270479b4c4686ace72bccaacff226


src/store/migrations/14_rsk_derivation_path.js
```diff
@@ -0,0 +1,75 @@
...
+          const updatedAccount = {
+            ...account,
+            alias: '',
+            addresses: [],
+            balances: {},
+            derivationPath
+          }
+          updatedAccounts.push(updatedAccount)
+
+          if (account.chain === ChainId.Rootstock) {
```
monokh, line 41: Will this add 2 accounts for ledger too? 
We need to decide what to do there, do you know other people using ledger, which path do they usually get? for example on metamask

Right now, we didn't migrate users to the new derivation path for ledger.

bradleySuira: Agree, fixed

src/store/getters.js
```diff
@@ -32,18 +33,25 @@ export default {
...
       const account = accountId ? getters.accountItem(accountId) : null
-      const accountType = account?.type || walletType
-      const accountIndex = account?.index || index
+      const _accountType = account?.type || accountType
+      const { chain } = cryptoassets[asset]
+      let derivationPath
+      if (account && account.derivationPath) {
+        derivationPath = account.derivationPath
+      } else {
+        derivationPath = getDerivationPath(chain, network, accountIndex, _accountType)
```
monokh, line 46: Hmm, why would an account ever not have a derivation path?

src/store/migrations/14_rsk_derivation_path.js
```diff
@@ -0,0 +1,75 @@
+import { getDerivationPath } from '@/utils/derivationPath'
+import { Networks } from '@/utils/networks'
+import { accountCreator, getNextAccountColor } from '@/utils/accounts'
+import { ChainId, chains } from '@liquality/cryptoassets'
+
+export const rskDerivationPath = {
```
monokh, line 6: Name of migration doesn't seem appropriate
shoudl be more like `account_set_derivation_path`

src/utils/derivationPath.js
```diff
@@ -0,0 +1,63 @@
...
+  if (accountType.includes('bitcoin_ledger')) {
+    const option = LEDGER_BITCOIN_OPTIONS.find(o => o.name === accountType)
+    const { addressType } = option
+    return `${BTC_ADDRESS_TYPE_TO_PREFIX[addressType]}'/${coinType}'/${index}'`
+  } else {
+    return `${BTC_ADDRESS_TYPE_TO_PREFIX[bitcoin.AddressType.BECH32]}'/${coinType}'/${index}'`
+  }
+}
+
+const getEthereumBasedDerivationPath = (coinType, index) => `m/44'/${coinType}'/${index}'/0/0`
```
monokh, line 17: `getEthereumBaseDerivationPath`?

src/utils/derivationPath.js
```diff
@@ -0,0 +1,63 @@
...
+    const ethNetwork = ChainNetworks[ChainId.Arbitrum][network]
+    return getEthereumBasedDerivationPath(ethNetwork.coinType, index)
+  },
+  [ChainId.Solana]: (network, index) => {
+    const solanaNetwork = ChainNetworks[ChainId.Solana][network]
+    return `m/44'/501'/${solanaNetwork.walletIndex}'/${index}'`
+  }
+}
+
+export const getDerivationPath = (chainId, network, index, accountType) => {
```
monokh, line 61: We should write some tests for this one. Easy to test, super critical and should almost never change

src/utils/accounts.js
```diff
@@ -1,11 +1,13 @@
...
 import { getAssetIcon } from '@/utils/asset'
+import { getDerivationPath } from '@/utils/derivationPath'
 import { chains } from '@liquality/cryptoassets'
 
 export const accountCreator = (payload) => {
-  const { walletId, account } = payload
+  const { network, walletId, account } = payload
   const {
     name,
+    alias,
```
monokh, line 10: What is the deal with this `alias` thing? I see it's been added to all the structures, but I can't see where it's being used.

src/views/Onboarding/OnboardingSetup.vue
```diff
@@ -63,7 +63,11 @@ export default {
     async confirmMnemonic () {
       this.currentStep = 'congrats'
       await this.setupWallet({ key: this.password })
-      await this.createWallet({ key: this.password, mnemonic: this.mnemonic }) // mnemonic prop can be null to generate new seed
+      await this.createWallet({
+        key: this.password,
+        mnemonic: this.mnemonic,
+        imported: !!this.imported
```
monokh, line 69: You could also just check that mnemonic is null as indicated in the comment below. 
Although i think `imported` prop is much cleaner

src/router/index.js
```diff
@@ -51,7 +51,7 @@ const routes = [
     component: UnlockWallet
   },
   {
-    path: '/onboarding/setup/:passphrase?',
+    path: '/onboarding/setup/:passphrase?/:passphrase?',
```
monokh, line 54: what does this mean? 😄  why passphrase twice?

monokh (changes requested):

src/store/getters.js
bradleySuira: hehe agree, but we can use the ledger provider we don't have it
the account comes from the accountId  const account = accountId ? getters.accountItem(accountId) : null, but can be null so when is null we just get the derivation path the case for this is when we want to get the accounts from ledger

src/store/migrations/14_rsk_derivation_path.js
bradleySuira: got it, fixed

src/utils/derivationPath.js
bradleySuira: should be re-used for all the EVM based chains

src/utils/accounts.js
bradleySuira: it's related to the structure that we set with ui/ux, for account names we will show something like:
`{name} - {alias}`, an example can be `Bictoin 1 - My user account name`

src/views/Onboarding/OnboardingSetup.vue
bradleySuira: Agreed with you fixed and checking now if we have the mnemonic or not and set the imported flag

src/router/index.js
bradleySuira: fixed, was a typo related and should be imported but we don't need it anymore

src/store/getters.js
bradleySuira: added a change to ask just when is ledger and we don't have an existing account

src/utils/derivationPath.js
bradleySuira: added

monokh (approved):

PR #254, opened 2021-09-01
feat: VUE_APP_AMPLITUDE_API_KEY

beemi:
## What?

-  Set the VUE_APP_AMPLITUDE_API_KEY as env
- Fix the Puppeter Docker image version with 1.0.6
- Rename the Release ci name


monokh (approved):
Lgtm
Just note that your key will only be set on Dev wallet the way you have added it I think

PR #256, opened 2021-09-02
fix: Fix the Network changed event

beemi:
## What?

💰 fix: Fix the Network changed event

💰  feat: Added Connect to Dapp event

### Changes

❌  -  Release CI config updated with set Amplitude API KEY against major tag releases, not for patch versions.

📚 -  DEV store extension doesn't send any events to Amplitude.

💻  - Added console log to check if Analytics enabled or not.

### Note
_The Dapp connection events only tracked if the user connected for the first time allowed, if the user comes back and try to connect dapp again without opening the allowed pop, in this case, we are not able to track that event _

### Next events

- Try to Track every connection with the same dapp?

image


src/background.js
```diff
@@ -28,8 +26,9 @@ store.subscribe(async ({
       store.dispatch('updateMarketData', { network: state.activeNetwork })
 
       store.dispatch('trackAnalytics', {
-        event: `Network Changed ${network} to ${state.activeNetwork}`
+        event: `Network Changed ${payload.currentNetwork} to ${payload.network}`
       })
+      console.log(this.state.activeNetwork)
```
monokh, line 31: console log required?

src/background.js
```diff
@@ -95,5 +94,16 @@ store.subscribe(async ({
         event: 'Lock Wallet'
       })
       break
+
+    case 'ADD_EXTERNAL_CONNECTION':
+      store.dispatch('trackAnalytics', {
+        event: 'Connect to Dapp',
+        properties: {
+          action: `${payload.chain} Dapp connected`,
```
monokh, line 102: i'm not familiar with standard for action,cateogry,label. but I assume these are in the correct format

src/components/Head.vue
```diff
@@ -50,7 +50,7 @@ export default {
       this.showNetworks = false
     },
     async switchNetwork (network) {
-      await this.changeActiveNetwork({ network })
+      await this.changeActiveNetwork({ currentNetwork: this.activeNetwork, network })
```
monokh, line 53: Why have to pass in `currentNetwork`?
You can retrieve current network before mutation in the action
```
export const changeActiveNetwork = async ({ commit, state }, {
  currentNetwork,
  network
}) => {
state.activeNetwork
```

.github/workflows/release.yml
```diff
@@ -42,24 +45,30 @@ jobs:
...
-      - name: Set development manifest
-        if: steps.semvers.outputs.v_patch == steps.get_current_tag.outputs.tag
-        run: |
-            jq --argjson icons "{\"16\": \"icons/dev/16x16.png\",\"48\": \"icons/dev/48x48.png\",\"128\": \"icons/dev/128x128.png\"}" '.icons = $icons | .browser_action.default_icon = $icons | .name = "Liquality Wallet - Dev"' ./src/manifest.json > ./src/manifest.tmp
-            mv ./src/manifest.tmp ./src/manifest.json
-
-      - name: Set development agents
-        if: steps.semvers.outputs.v_patch == steps.get_current_tag.outputs.tag
+    # Set the Amplitude Analytics only for Major prod releases.
+      - name: Set development agents (major for Prod)
```
monokh, line 49: The name seems wrong? Isn't this enabling analytics?

.github/workflows/release.yml
```diff
@@ -42,24 +45,30 @@ jobs:
...
+        if: steps.semvers.outputs.v_major == steps.get_current_tag.outputs.tag
```
monokh, line 50: Shouldn't it be `v_minor` ?
`major.minor.patch`
we release to prod on every `minor`

.github/workflows/release.yml
```diff
@@ -42,24 +45,30 @@ jobs:
...
         env:
           VUE_APP_AMPLITUDE_API_KEY: ${{ secrets.VUE_APP_AMPLITUDE_API_KEY }}
         run: |
             echo VUE_APP_AMPLITUDE_API_KEY =${{ secrets.VUE_APP_AMPLITUDE_API_KEY }} >> .env
-            sed -i "s/\/swap-testnet\/agent/\/swap-testnet-dev\/agent/g; s/\/swap\/agent/\/swap-dev\/agent/g" src/build.config.js
       - run: |
```
monokh, line 55: Since you are repeating this with the below job, you can just put this as a separate step to deduplicate

monokh (changes requested):

src/components/Head.vue
bradleySuira: changed to get this value from the state inside the action

.github/workflows/release.yml
monokh: oh not the sed 😄 I mean the build step `npm run ci, npm run lint, npm run build`

src/store/actions/changeActiveNetwork.js
```diff
@@ -1,5 +1,10 @@
-export const changeActiveNetwork = async ({ commit }, { network }) => {
+export const changeActiveNetwork = async ({ state, commit }, {
+  currentNetwork,
```
monokh, line 2: unused variable?

PR #260, opened 2021-09-06
feat: Send / Hold assets events

beemi:
- Send analytics
- Hold assets event after user selects the chain if balance is greater than 0
- Add Custom token events
- Rename other events / actions

https://www.notion.so/Implement-Wallet-Analytics-51e263d76eae4edaaaa5821ef98baef1

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -93,8 +93,7 @@ store.subscribe(async ({
         event: 'SEND',
         properties: {
           action: `Send Created on ${state.activeNetwork}`,
-          category: `Send to ${payload.transaction.to}`,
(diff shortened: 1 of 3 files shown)


bradleySuira (approved):
LGTM

src/background.js
```diff
@@ -75,34 +74,59 @@ store.subscribe(async ({
...
+      })
+
+      break
+
+    case 'NEW_TRASACTION':
+
+      store.dispatch('trackAnalytics', {
+        event: 'Send',
+        properties: {
+          action: 'Funds sent',
```
monokh, line 94: missing category, intended?

beemi: Fixed this the new PR

PR #263, opened 2021-09-07
feat: Refresh balance & Copy address from Receive screen events added

beemi:
## What?

- Receive screen copy address event
- Refresh balance on-chain screen event
- Swap update status
- Send update status
- Send event renamed
- Added DEV token for Amplitude for Dev extension

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,10 @@ jobs:
     # Build the extension for DEV store.
       - name: Set development agents(patch for dev)
         if: steps.semvers.outputs.v_patch == steps.get_current_tag.outputs.tag
+        env:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         with:
           version: ${{ steps.previous_tag.outputs.tag }}
 
-    # Enable Amplitude Analytics only for minor prod releases.
+    # Minor PROD releases.
       - name: Enable Analytics & Set development agents (minor for Prod)
         if: steps.semvers.outputs.v_minor == steps.get_current_tag.outputs.tag
```
bradleySuira, line 47: Not sure because we use v_major for PROD only and then v_patch for DEV

bradleySuira (changes requested):

PR #281, opened 2021-09-16
fix: pass through accountId to permission requests

monokh:
## What?

Instead of retrieving the currently connected account when the request is being executed,
send the accountId through the permission request. This ensures the request is locked to the correct account when it is accepted and also allows us to display account information in the permission requests in the future

## Testing?
Test injection with dapps (regression)
I tested sovryn continues to work and connect to correct account


bradleySuira (approved):
LGTM

PR #289, opened 2021-09-19
feat: update balance event

beemi:
## What?

image

### DIFF
--- a/src/views/Account.vue
+++ b/src/views/Account.vue
@@ -229,17 +229,6 @@ export default {
         label: `Select ${this.asset} (${chain})`
       }
     })
-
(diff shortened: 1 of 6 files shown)


src/views/Account.vue
```diff
@@ -232,11 +232,11 @@ export default {
 
     if (this.balance > 0) {
```
monokh, line 233: Why 2 places for this?

beemi: - One at when use click on actual asset - Ex: Click on BTC 
- One at as soon as they landed on overview page - Update balance

monokh (approved):

PR #300, opened 2021-09-28
fix: show token symbol on allowance

skeremidchiev:
## What?
Properly present token symbol (RSK network) when allowance is signed.

#Notion/ Trello
https://www.notion.so/30ca4d49f3844a0fabdf4141f8a8bccc?v=339d1a8147b84a01a1eeca3e04eecdd7&p=d50cc69818aa456a887b3fdcd5fd34a0
Fix is related to issue 1.

## Why?
Better user experience. 

## Testing?
- [x] Run tests locally

## Screenshots (optional)

Screenshot 2021-09-28 at 10 21 23

## Anything Else?
This change could have a significant impact on other dapps. It should be tested thoroughly!


src/views/PermissionSend.vue
```diff
@@ -144,6 +144,13 @@ export default {
         } catch {
           this.symbol = 'ETH'
         }
+      } else if (this.assetChain === 'RBTC') {
```
monokh, line 147: Seems like we are repeating here. 
Why don't we get the token details regardless of what chain? `tokenDetailProviders` allows that

But in either case, I don't think we should be retrieving token details from the chain like this. Otherwise a dapp could trick a user by asking them to approve a fake token that looks like "DAI" or something else

Only if the token address is listed in `cryptoassets` lib should we display the name

monokh (changes requested):

PR #330, opened 2021-10-13
fix: Correctly deny permission to request. requesting again works

monokh:
#Linear Ticket

Fixes LIQ-89


beemi:
@monokh  i think it was related to https://linear.app/liquality/issue/LIQ-312/[dapp-multiaccount]-permission-was-not-given-while-injection

PR #334, opened 2021-10-14
feat: Update min amount for FastBTC to 0.0004

AndonMitev:
Task: https://linear.app/liquality/issue/LIQ-111/users-reporting-sender-account-does-not-exist:

image


AndonMitev:
> This doesn't really behave the same way the ticket describes
> 
> In this case, min is only changed when selecting fastBTC as provider first
> 
> In the ticket, it describes the min should always be 0.0004, and that the default amount should still be liquality's min
> 
> What do we want to do?

AndonMitev:
> This doesn't really behave the same way the ticket describes
> 
> In this case, min is only changed when selecting fastBTC as provider first
> 
> In the ticket, it describes the min should always be 0.0004, and that the default amount should still be liquality's min
> 
> What do we want to do?

@monokh The task were updated to pick the lowest minimum amount between multiple providers

skeremidchiev (approved):

PR #353, opened 2021-10-26
feat: fuse integration into liquality wallet (WIP)

t0mcr8se:
Fuse integration into liquality wallet.

## Instructions:
- Refer to the issues section below before testing.
- Get some testnet fuse here 👉 https://get.fusespark.io
- Load the extention
- Add fuse from Manage assets list

## Tasks:
These are the steps provided in the Liquality Integration Guide V1 (step III) 
- [x] Add an svg for the new chain
- [x] Add the symbol of the new chain
- [ ] ~~Add a provider of the new chain here: src/contentScript.js~~ (not needed)
- [ ] ~~Add new provider following the logic for the other chains here: src/inject.js~~ (not needed)
- [x] Add all implemented providers for the given chain in step 1
- [x] Add the Symbol of the new chain in the getters
- [x] Add all supported networks for the given chain: src/utils/asset.js
- [x] Add some fixed fee values if available: src/utils/fees.js 
- [x] Add faucet url if available: src/views/Receive.vue 

## Result:
We can add the fuse network and see the correct address generated

## Issues
Requests from liquality seems to omit the `"jsonrpc":"2.0"` in the request body, and we get this error: "Unsupported JSON-RPC protocol version".

This is a common error in OpenEthereum nodes when the jsonrpc version is not specified in the request, and our nodes run on OpenEthereum.


src/build.config.js
```diff
@@ -22,7 +22,8 @@ export default {
       'ARBETH',
       'FISH',
       'LUNA',
-      'UST'
+      'UST',
+      'FUSE'
```
monokh, line 26: Please remove fuse from the default assets,
only assets which we have vetted very closely are added to the user's wallet by default
Users can still use fuse by enabling it in Settings -> Manage assets

src/store/factory/client.js
```diff
@@ -216,6 +216,18 @@ function createArbitrumClient (asset, network, mnemonic, derivationPath) {
   return createEthereumClient(asset, network, arbitrumNetwork, rpcApi, scraperApi, feeProvider, mnemonic, 'default', derivationPath)
 }
 
+function createFuseClient (asset, network, mnemonic, derivationPath) {
+  const isTestnet = network === 'testnet'
+  const fuseNetwork = ChainNetworks.fuse[network]
+  const rpcApi = isTestnet ? 'https://fuse-testnet.gateway.pokt.network/' : 'https://rpc.fuse.io'
+  const scraperApi = isTestnet ? 'https://liquality.io/polygon-testnet-api' : 'https://liquality.io/polygon-mainnet-api' // TODO: Change this
```
monokh, line 223: scraper api not required

monokh (changes requested):
Please provide some instructions to test
where to get testnet funds
dapps

t0mcr8se:
@monokh Just updated the description above, please have a look.

PR #363, opened 2021-11-01
feat: Add erc20 to terra

AndonMitev:
LIQ-297

https://linear.app/liquality/issue/LIQ-297/need-to-add-terra-chain-option-to-add-custom-token-under-manage-assets


src/store/factory/client.js
```diff
@@ -217,20 +217,43 @@ function createArbitrumClient (asset, network, mnemonic, derivationPath) {
...
+    } case 'UST': {
+      _asset = 'uusd'
+      feeAsset = 'uusd'
+      terraNetwork = { ...ChainNetworks.terra[network] }
+      break
+    } default: {
+      _asset = asset
+      feeAsset = 'uluna'
+      tokenAddress = cryptoassets[asset].contractAddress
+      terraNetwork = { ...ChainNetworks.terra[network] }
```
monokh, line 237: Why is this spreading the object ? 😄

src/store/state.js
```diff
@@ -51,5 +51,6 @@ export default {
   },
   experiments: {},
   watsNewModalVersion: null,
-  enabledChains: {}
+  enabledChains: {},
+  terraToken: {}
```
monokh, line 55: Why do we still have this state and the mutation ?

src/utils/asset.js
```diff
@@ -228,3 +230,16 @@ export const estimateGas = async ({ data, to, value }) => {
 
   return await provider.estimateGas(paramsForGasEstimate)
 }
+
+export const fetchTerraToken = async (address, network) => {
+  const { data: { [network]: tokens } } = await axios.get('https://assets.terra.money/cw20/tokens.json')
+  const token = tokens[address]
+  const { symbol, token: tokenAddress } = token
+  store.commit('TERRA_TOKEN', { symbol, tokenAddress })
```
monokh, line 238: I don't think we need this state anymore right? Everything is stored fine on `customTokens`

monokh: And can't this go in `tokenDetailProviders` like the rest of them?

monokh: I see there are some slightly different requirements here compared to the others but we should just make the interface compatible

I think it will work if all the `getDetails` functions take both address and network

src/store/mutations.js
```diff
@@ -330,5 +330,11 @@ export default {
 
       Vue.set(state.accounts[walletId][network], index, updatedAccount)
     }
+  },
+  TERRA_TOKEN (state, { symbol, tokenAddress }) {
```
monokh, line 334: This is still here? :S

monokh (approved):

monokh (approved):

PR #365, opened 2021-11-01
feat: experiments panel

b0b4456:
https://linear.app/liquality/issue/LIQ-425/prototype-analytics-and-experimental-mode-opt-in-in-wallet

## Why?
internal experiments in prod

## How?
click the version number in the settings to potentially access more settings

## Screenshots 
capture


monokh (approved):
👍  LGTMTYWAGI

PR #367, opened 2021-11-03
feat: Add computed prop which contains all of the prefixes needed for the QR code

AndonMitev:
feat: Add computed prop which contains all of the prefixes needed for the QR code

#Linear Ticket

LIQ-610

### DIFF
--- a/src/views/Receive.vue
+++ b/src/views/Receive.vue
@@ -98,6 +98,18 @@ export default {
     routeSource () {
       return this.$route.query.source || null
     },
+    prefix() {
+      return ({
+        bitcoin: 'bitcoin:',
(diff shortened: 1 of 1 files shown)


src/views/Receive.vue
```diff
@@ -98,6 +98,18 @@ export default {
     routeSource () {
       return this.$route.query.source || null
     },
+    prefix() {
+      return ({
```
monokh, line 102: Can it reuse the `chainName` property? 
Just concious of adding more places that we need to update for integrations :D

monokh (approved):

monokh:
updated in another pr

PR #378, opened 2021-11-09
fix: min value for liqualityboost and minimum default swap value

tima-t:
LIQ-609


monokh (changes requested):
It's going to be difficult to maintain the default list because assets are added and values change. 

Have you seen the logic that sets the default amount for swaps?

https://github.com/liquality/wallet/blob/d61875595b25452a5be5977cef6ce64c98122fe1/src/views/Swap/Swap.vue#L545

It uses $100 as a default value when a liquality market does not exist

Why did this mechanism not work to fix this case?

bradleySuira (approved):
LGTM

tima-t:
> 

I agree with what you are saying. 

The thing is if you have 1inch swap provider on Matic or BSC what is the minimum? 
We can set $100 by default, but maybe this is not precise, that why I decided to put something like 2$. 

The other logic you are referring to is about default value and this is about minimum value.
 I agree that my logic is not very scalable as we have to update these rates as they are dynamic. 
 
 However, if we agree on that - we can put $100 as minimum value  and use this kind of logic `BN.min(fiatToCrypto(DEFAULT_SWAP_VALUE_USD, this.fiatRates[this.asset])`

beemi (approved):

PR #380, opened 2021-11-09
fix: Calculate and deduct tax for sending max ust

AndonMitev:


src/views/UnlockWallet.vue
```diff
@@ -52,7 +52,7 @@ export default {
     return {
       loading: false,
       error: null,
-      password: ''
+      password: 'andon940410'
```
monokh, line 55: Ser

src/views/Swap/Swap.vue
```diff
@@ -792,6 +799,18 @@ export default {
         for (const [speed, fee] of Object.entries(assetFees)) {
           fees[chain][speed] = fees[chain][speed].plus(totalFees[fee.fee])
         }
+
+        if (asset === 'UST' && chain === 'LUNA') {
```
monokh, line 803: Shouldn't be handled differently to bitcoin... should be implemented in `estimateFees` in `LiqualitySwapProvider`
One issue is that you don't have balance there, 
you can either pass it or I think better is if `getTaxFees` is on `TerraWalletProvider` and can check the balance internally. This way the parameters passed will be very similar to bitcoin

src/views/Swap/Swap.vue
```diff
@@ -563,7 +564,13 @@ export default {
       return dpUI(min)
     },
     max () {
-      return this.available && !isNaN(this.available) ? BN.min(BN(this.available)) : BN(0)
+      const max = this.available && !isNaN(this.available) ? BN.min(BN(this.available)) : BN(0)
+
+      if (max && this.asset === 'UST') {
```
monokh, line 569: This shouldn't be required if `LiqualitySwapProvider.estimateFees` is implemented correctly

monokh (changes requested):

monokh (approved):

PR #385, opened 2021-11-10
fix: thorchain swap

tima-t:
Fixes LIQ-584


src/views/Swap/Swap.vue
```diff
@@ -572,6 +572,8 @@ export default {
       return fee || BN(0)
     },
     toSwapFee () {
+      if (this.selectedQuote?.toSwapFees) return unitToCurrency(cryptoassets[this.toAsset], this.selectedQuote.toSwapFees).toFixed() // Thorchain case workaround
```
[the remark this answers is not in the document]
monokh: you can define `toTxType` in `ThorchainSwapProvider`
and then `estimateFees` will be called with the parameters where you can refund the `toSwapfees` calculation

src/views/Swap/Swap.vue
```diff
@@ -626,15 +628,19 @@ export default {
         return 'Please increase amount. It is below minimum.'
       }
 
+      if (this.selectedQuote?.maxFeePercentage && (BN(this.selectedQuote.toAmount).times(this.selectedQuote?.maxFeePercentage)).lt(this.selectedQuote.toSwapFees)) {
```
monokh, line 631: I don't think we need this, there is already a mechanism during swap review page that tells you if the swap you are making makes sense considering the fee
e.g. : 
image

src/views/Swap/Swap.vue
```diff
@@ -665,11 +669,12 @@ export default {
       return send.plus(fee).toFormat(2)
     },
     receiveAmountSameAsset () {
-      return BN(this.receiveAmount).minus(this.toSwapFee)
+      return BN(this.receiveAmount).minus(this.toSwapFee).toFixed()
     },
     totalToReceiveInFiat () {
       const receive = cryptoToFiat(this.receiveAmount, this.fiatRates[this.toAsset])
-      const fee = cryptoToFiat(this.toSwapFee, this.fiatRates[this.toAssetChain])
+      const fee = cryptoToFiat(this.toSwapFee, this.selectedQuote?.toSwapFees ? this.fiatRates[this.toAsset] : this.fiatRates[this.toAssetChain]) // Thorchain case workaround
```
monokh, line 676: again, I think workarounds like this are not desired.

monokh (changes requested):

src/views/Swap/Swap.vue
tima-t: We had a talk with Simon and Alok and we decided to do it like this. 
If the fee is more than 15% of the received amount to show a message and block the swap.

src/swaps/views/Timeline.vue
```diff
@@ -374,7 +374,9 @@ export default {
         : this.getTransactionStep(completed, pending, side, this.item.swapTxHash, this.item.swapTxHash, this.item.to, 'swap')
     },
     async getReceiveStep (completed, pending, side) {
-      return this.getTransactionStep(completed, pending, side, this.item.receiveTxHash, null, this.item.to, 'receive')
+      return this.item.status === 'FAILED'
```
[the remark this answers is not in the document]
tima-t: Yes, that makes sense.

src/views/Swap/Swap.vue
monokh: I understand but how is that different from the mechanism in the screenshot?

I mean we shouldn't have 2 different ways of doing the same thing. If we want to block swap from happening when fee is too high, we should define a specific threshold and enforce it for all swaps the same way

src/views/Swap/Swap.vue
```diff
@@ -633,6 +633,10 @@ export default {
         return 'Please increase amount. It is below minimum.'
       }
 
+      if (this.selectedQuote?.receiveFee && BN(this.selectedQuote.toAmount).lt(this.selectedQuote.receiveFee)) {
+        return 'Increase amount. It won`t cover receive fee.'
```
harshjv, line 637: Can you change this line to

```
"Increase amount. It won't cover receive fee."
```

harshjv (changes requested):

PR #392, opened 2021-11-16
feat: migrate 1inch to v4

tima-t:
LIQ-685


bradleySuira (approved):
LGTM

src/store/getters.js
```diff
@@ -43,6 +43,10 @@ const TESTNET_ASSETS = [
   })
 }, {})
 
+const mapLegacyProvidersToSupported = {
+  oneinchV3: 'oneinchV4'
```
kraikov, line 47: Do we need `oneinchV3` at all now? Is `oneinchV4` supposed to completely replace `oneinchV3` ?

src/store/getters.js
```diff
@@ -89,12 +93,13 @@ export default {
   },
   swapProvider (state) {
     return (network, providerId) => {
-      const cacheKey = [network, providerId]
+      const supportedProviderId = mapLegacyProvidersToSupported[providerId] ? mapLegacyProvidersToSupported[providerId] : providerId
```
kraikov, line 96: Why is this needed now?

src/store/getters.js
tima-t: Yes, they are doing the same and V3 will be deprecated soon . 

I need this line, because we have swaps in the history that have been using V3 and they are in the local storage. 
If we do not have a backward support this will result in visualization issues in the old swaps

src/store/getters.js
tima-t: Yes, they are doing the same and V3 will be deprecated soon . 

I need this line, because we have swaps in the history that have been using V3 and they are in the local storage. 
If we do not have a backward support this will result in visualization issues in the old swaps. 

I can also leave the old oneinchV3 and supported v3 and v4, but it does not make much sense to me

skeremidchiev (approved):

PR #395, opened 2021-11-16
fix: fetch missing rsk token prices from the Sovryn contracts

kraikov:
This should fix:
- [LIQ-417](https://linear.app/liquality/issue/LIQ-417/sovryn-amm-provider-receiving-rates-are-incorrect)
- [LIQ-129](https://linear.app/liquality/issue/LIQ-129/[dollar-price-difference-between-wallets]-sovryn-vs-liquality)
- [LIQ-45](https://linear.app/liquality/issue/LIQ-45/we-need-usd-prices-for-custom-assets-beyond-coingecko-api) **partially**

The fix is not ideal, but the way prices are fetched has to be changed/refactored in general in order to implement that in a better way.

Fixes LIQ-417
Fixes LIQ-129
Fixes LIQ-45


monokh:
I believe we decided against this approach

PR #397, opened 2021-11-17
feat: updated rsk endpoints to use sovryn endpoints

bradleySuira:
#Linear Ticket

[https://linear.app/liquality/issue/LIQ-123/evaluate-moving-to-sovryns-endpoints-503-node-error](https://linear.app/liquality/issue/LIQ-123/evaluate-moving-to-sovryns-endpoints-503-node-error)


beemi (approved):
LGTM

src/store/factory/client.js
```diff
@@ -177,7 +177,7 @@ function createSolanaClient (network, mnemonic, derivationPath) {
 function createRskClient (asset, network, mnemonic, accountType, derivationPath) {
   const isTestnet = network === 'testnet'
   const rskNetwork = ChainNetworks.rsk[network]
-  const rpcApi = isTestnet ? 'https://public-node.testnet.rsk.co' : 'https://public-node.rsk.co'
+  const rpcApi = isTestnet ? 'https://testnet.sovryn.app/rpc' : 'https://mainnet.sovryn.app/rpc'
```
monokh, line 180: There are also other occurances of this in the wallet, please update / consolidate

monokh (changes requested):

src/store/factory/client.js
bradleySuira: got it, let me check right now
Thanks!

bradleySuira: @monokh done, it's fixed now, added an env var to set mainnet and testnet endpoints then if we need to update in the future only the env var should be updated

beemi (approved):

PR #398, opened 2021-11-18
fix: swap and send input issue - `NaN`

Jennievon:
#LIQ-652 && #LIQ-34


src/views/Send/Send.vue
```diff
@@ -293,7 +306,7 @@ export default {
     },
     amountFiat: {
       get () {
-        return `$${this.stateAmountFiat}`
+        return this.stateAmountFiat
       },
       set (newValue) {
         const value = (newValue || '0').replace('$', '')
```
bradleySuira, line 312: I think that we don't need the replace if we are using input type number and you removed the $ sign from the getter

src/views/Send/SendInput.vue
```diff
@@ -82,7 +84,7 @@
       </div>
       <div class="send-bottom-available" id="send_available_balance">
         <span class="text-muted">Available</span>
-        {{ isNaN(available) ? '0' : dpUI(available) || '0' }} {{ asset }}
+        {{ isNaN(available) ? "0" : dpUI(available) || "0" }} {{ asset }}
```
bradleySuira, line 87: we are using single quotes

src/views/Swap/SendInput.vue
```diff
@@ -105,7 +106,7 @@
       </div>
       <div class="swap-send-bottom-available">
         <span class="text-muted">Available</span>
-        {{ isNaN(available) ? '0' : dpUI(available) || '0' }} {{ asset }}
+        {{ isNaN(available) ? "0" : dpUI(available) || "0" }} {{ asset }}
```
bradleySuira, line 109: single quotes

bradleySuira (changes requested):

bradleySuira (approved):
LGTM

PR #401, opened 2021-11-19
feat: Added BTC export Private key test

beemi:
# Linear Ticket
LIQ-23
https://linear.app/liquality/issue/LIQ-23/develop-ability-to-export-private-keys-for-a-given-account


b0b4456 (approved):
👍

tests/e2e/19_private_keys.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    await page.waitForSelector('#i_have_privacy_button', { visible: true })
+    await page.click('#i_have_privacy_button')
+    await page.waitForSelector('#password', { visible: true })
+    await page.type('#password', password)
+    await page.click('#checkbox')
+    await page.click('#continue_button_to_see_seed_phrase')
+    // Private key screen
+    await page.waitForSelector('#private-key-textarea', { visible: true })
+    const privateKeyTextArea = await page.$eval('#private-key-textarea', el => el.getAttribute('readonly'))
+    expect(privateKeyTextArea).equals('readonly')
```
b0b4456, line 59: maybe also expect that the textarea contains a string that looks like a privatekey, eg something other than `n/a` ?

beemi: yes trying to get the textContent but its empty
image

beemi: ` const privateKey = await page.$eval('#private-key-textarea', el => el.value)
    expect(privateKeyTextArea).equals('readonly')
    expect(privateKey, 'Private key export value shouldn\'t be n/a')
      .to.not.contain.oneOf(['n/a', 'NaN', null])`

beemi: image

PR #414, opened 2021-11-29
refactor: LIQ-704

Jennievon:
- update address display
- update copy tooltip

#[LIQ-704](https://linear.app/liquality/issue/LIQ-704/update-all-shortened-account-addresses-with-[6-digits][4digits]-and)


bradleySuira (changes requested):
Hi @Jennievon changes are really good but a little issue with ui when mouse hover, can you check please ? 
Screen Shot 2021-11-29 at 10 34 15 AM

If you want please ask to @Thessy or Vera but I think that we need to move the eye icon to the right

Jennievon:
Hi @bradleySuira, 
I have it on my PC as this.

Screenshot 2021-11-29 at 19 52 04

bradleySuira:
@Jennievon I think that it works well for BTC but for ETH (check your Ethereum account) or EVM chains we use `0x` prefix and it's the issue

Jennievon:
@bradleySuira do you think we should use substring of 0,4 for addresses with prefix?

bradleySuira:
Hi @Jennievon I this case we need to keep the 0x prefix for all EVM based blockchains so the fix can be using some css, ping me at discord if you want

BR

src/utils/address.js
```diff
@@ -1,6 +1,6 @@
 export function shortenAddress (address) {
   const prefix = address.startsWith('0x') ? '0x' : ''
-  return `${prefix}${address.replace('0x', '').substring(0, 4)}...${address.substring(address.length - 4)}`
+  return `${prefix}${address.replace('0x', '').substring(0, prefix ? 6 : 4)}...${address.substring(address.length - 4)}`
```
Jennievon, line 3: @bradleySuira I did this check here.

bradleySuira: LGTM

bradleySuira (approved):
LGTM

bradleySuira (approved):
LGTM

src/utils/address.js
kraikov: Is this used for all addresses? I'm asking because the prefix for other chains could be different. For example in Terra we have is `terra1...`

Jennievon: Hi @kraikov, so in the case where there is a prefix, we're leaving it at 4.

beemi:
Ok this PR will be in Draft mode until we resolve the terra chain and other testnet address

bradleySuira (approved):
LGTM

beemi (approved):

PR #416, opened 2021-11-30
fix: Update terra swap gas limit

AndonMitev:
Update terra base gas limit from 650000 to 800000


kraikov (approved):
LGTM.

The magic number 800000 is taken from Terra station wallet.

https://github.com/terra-money/station/blob/d99b250a458c59d786cc7831aa6961adcd735474/src/post/hooks/wallet/useSwap.ts#L256

beemi (approved):

PR #417, opened 2021-11-30
feat: changelog api

bradleySuira:

#[LIQ-812](https://linear.app/liquality/issue/LIQ-812/implement-new-changelog-api-into-the-wallet)


monokh:
Open again if work is to continue

PR #422, opened 2021-12-03
feat: Improve UX for initial loading of overview page

AndonMitev:
# :books: https://linear.app/liquality/issue/LIQ-131/[loading-issue]-taking-more-than-3-4-minutes-to-load-total-assets :books:

Changes:
1. Remove global loading in order to display immediately Send/Swap/Receive buttons to the user.
2. Add loading against every chain and this loading is displayed until the balance is fetched

For more details check linear ticket comment section


src/views/Wallet/WalletStats.vue
```diff
@@ -60,7 +59,6 @@ export default {
     ReceiveIcon,
     SwapIcon
   },
-  props: ['loading'],
```
bradleySuira, line 63: I think that we still need the loading indicator but only to show the total fiat balance when all the accounts balances loaded. Tested locally but the total fiat balance seems to be not updated when all balances are loaded

Screen Shot 2021-12-03 at 10 32 11 AM

bradleySuira (changes requested):

src/views/Wallet/WalletStats.vue
AndonMitev: Hmm that weird, so we need to put back loading indicator for total fiat balance and display only action buttons?
image

AndonMitev: @bradleySuira can you check if this behaviour is reproducible only for ledger because i don't have one to test that case ;s

bradleySuira (approved):
LGTM

kraikov (approved):
LGTM

tima-t (approved):
LGTM

PR #431, opened 2021-12-07
feat: Prepend network if symbol already exists

AndonMitev:
# :books: https://linear.app/liquality/issue/LIQ-578/add-message-for-editing-token-symbol-luna-while-adding-the-token-for :books:


beemi (approved):

src/views/CustomToken.vue
```diff
@@ -136,11 +136,18 @@ export default {
     networkAssets () {
       return this.enabledAssets[this.activeNetwork][this.activeWalletId]
     },
-    isExistingNetworkAsset () {
-      return Boolean(this.networkAssets.find(_symbol => _symbol === this.symbol))
+    chainNativeAsset() {
+      return {
+        ethereum: 'ETH',
```
skeremidchiev, line 141: It's a better approach to get chains from cryptoassets.

AndonMitev: fixed

skeremidchiev (approved):

src/views/CustomToken.vue
```diff
@@ -214,8 +218,9 @@ export default {
         this.isSymbolEditable = Boolean(this.networkAssets.find(_symbol => _symbol === this.symbol))
       }
     }, 500),
-    async selectChain (chain) {
+    async selectChain (chain, nativeAsset) {
       this.chain = chain
+      this.nativeAsset = nativeAsset
```
bradleySuira, line 223: Hi, I think that we don't need to set the nativeAsset manually, we can get it from cryptoassets:
```
import { chains } from '@liquality/cryptoassets'
...
async selectChain (chain) {
 this.chain = chain
 this.nativeAsset = nativeAsset
 const { nativeAsset } = chains[nativeAsset]
 this.nativeAsset = nativeAsset
...
}
```

bradleySuira (changes requested):

beemi:
@AndonMitev  can u please last comment

PR #493, opened 2021-12-15
fix: resolve migration issue

skeremidchiev:

[LIQ-885](https://linear.app/liquality/issue/LIQ-885/migration-failing)

It seems like currently migration is not properly working for more than a single version increase.


src/store/migrations/index.js
```diff
@@ -53,7 +53,7 @@ async function processMigrations (state) {
   for (const migration of migrations) {
     if (currentVersion < migration.version) {
       try {
-        newState = await migration.migrate(cloneDeep(state))
+        newState = await migration.migrate(cloneDeep(newState))
```
harshjv, line 56: actually, already `newState = cloneDeep(state)`. do we need to `deepClone` it twice?

harshjv (changes requested):

src/store/migrations/index.js
skeremidchiev: It seems like we do because of this case:
1) start migration on top of non deep copied state.
2) migration fails midway
3) exception is caught and code breaks out of the loop
4) method returns `half backed` state

monokh (approved):

src/store/migrations/index.js
kraikov: I agree with @skeremidchiev here.

The way I see this is:

```js
currentVersion = 10
latest migration version = 15
```

Then we have:
````js
newState = [...migration11, ...newState]
newState = [...migration12, ...newState]
...
newState = [...migration15, ...newState]

kraikov (approved):
LGTM

harshjv (changes requested):

src/store/migrations/index.js
kraikov: `cloneDeep` returns the deep cloned value, so `newState` cannot be changed within the migration

skeremidchiev: Well, I don't see a reason to make it immutable, constant or deepFreeze-ing it. `cloneDeep` is passing a brand new copy of the state which prevents messing up the object.

monokh (approved):

PR #497, opened 2021-12-16
fix: do not show rates when they are missing

tima-t:
# 📚   liq-875 📚

### DIFF
--- a/src/utils/coinFormatter.js
+++ b/src/utils/coinFormatter.js
@@ -24,13 +24,13 @@ export const prettyBalance = (amount, coin, dp = VALUE_DECIMALS) => {
 }
 
 export const prettyFiatBalance = (amount, rate) => {
-  if (!amount || !rate) return amount
   const fiatAmount = cryptoToFiat(amount, rate)
+  if (isNaN(fiatAmount)) return fiatAmount
   return formatFiat(fiatAmount)
 }
 
 export const cryptoToFiat = (amount, rate) => {
-  if (!rate) return BN(amount)
+  if (!rate) return 'Rate missing'
   return BN(amount).times(rate)
 }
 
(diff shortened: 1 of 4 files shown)


src/views/Swap/ReceiveInput.vue
```diff
@@ -84,7 +84,7 @@ export default {
   created () {},
   computed: {
     receiveAmountFiatValue () {
-      return '$' + dpUI(this.receiveAmountFiat, 2)
+      return isNaN(this.receiveAmountFiat) ? this.receiveAmountFiat : ('$' + dpUI(this.receiveAmountFiat, 2))
```
skeremidchiev, line 87: Add space between `$` and the value.

skeremidchiev (approved):

src/utils/coinFormatter.js
```diff
@@ -24,13 +24,13 @@ export const prettyBalance = (amount, coin, dp = VALUE_DECIMALS) => {
...
 export const prettyFiatBalance = (amount, rate) => {
-  if (!amount || !rate) return amount
   const fiatAmount = cryptoToFiat(amount, rate)
+  if (isNaN(fiatAmount)) return fiatAmount
   return formatFiat(fiatAmount)
 }
 
 export const cryptoToFiat = (amount, rate) => {
-  if (!rate) return BN(amount)
+  if (!rate) return 'Rate missing'
```
skeremidchiev, line 33: A tip from @kraikov. Replace this line with:
```
try {
   BigNumber(rate) 
} catch(err) { 
   return ''
}
```

tima-t: I think this is good, but it is not consistent with the other validations in  the file. 
Should we refactor all of them or just leave it as it is now with validation 

if (!rate) return 'Rate missing'

beemi (approved):

b0b4456 (approved):

PR #519, opened 2021-12-23
feat: rework send.vue and swap.vue to both match new design

skeremidchiev:
[LIQ-121](https://linear.app/liquality/issue/LIQ-121/in-case-of-rsk-send-user-set-custom-fees-it-displays-in-scientific)


skeremidchiev:
PR is not completed due to:
1) an issue with switching between asset and USD value
2) not able to clean the input field once USD is set

bradleySuira (approved):
LGTM

beemi (approved):

PR #541, opened 2022-01-09
feat: integrate avalanche in wallet

tima-t:
📚 liq-800 📚


src/store/getters.js
```diff
@@ -57,7 +58,7 @@ export default {
       const account = accountId ? getters.accountItem(accountId) : null
       const _accountType = account?.type || accountType
       const _accountIndex = account?.index || accountIndex
-      const { chain } = getters.cryptoassets[asset]
+      const { chain } = getters.cryptoassets[asset] || cryptoassets[asset]
```
kraikov, line 61: Why is that required?

test/e2e/07_add_custom_token.spec.js
```diff
@@ -38,6 +38,13 @@ const customTokensDetails = [
     symbol: 'SUSHI',
     decimal: '18'
   },
+  {
+    chain: 'avalanche',
+    address: '0x0b9d5d9136855f6fec3c0993fee6e9ce8a297846',
+    name: 'LINK',
```
monokh, line 44: name and symbol :D

src/utils/cryptoassets.js
```diff
@@ -1,12 +1,13 @@
 import store from '../store'
+import { assets } from '@liquality/cryptoassets'
 
 // Redefine the `cryptoassets` lib to pull from the getter - to include custom tokens
 const cryptoassets = new Proxy({}, {
   get (target, name, receiver) {
-    return Reflect.get(store.getters.cryptoassets, name, receiver)
+    return Reflect.get({ ...store.getters.cryptoassets, ...assets }, name, receiver)
```
monokh, line 7: What is going on here? It looks like you are trying to fix a bug but It's not clear what it is or how to test it. Maybe remove this change from this PR and have another one so its clearer?

monokh (changes requested):

src/store/getters.js
tima-t: const { chain } = getters.cryptoassets[asset]  - takes all the assets that we have currenlty in the walleet cache, but if we add new assets in @liquality/cryptoassets package - they are not going to be detected.  

That's why the change is required

test/e2e/07_add_custom_token.spec.js
tima-t: ohhh sorry - will change it :)

src/utils/cryptoassets.js
tima-t: Yeah, I agree - I will create a new pr for this. 

The problem is that it is getting the tokens from the wallet store, but it do no detect the changes and new token coming from assets so we have to merge both stores to get the new assets.

src/store/getters.js
tima-t: I've removed this and moved it to another PR

test/e2e/07_add_custom_token.spec.js
tima-t: I fixed that - basically this is for the test cases and this is the token Link on avalanche, but now I added another token Link.e - as this is its symbol on avalanche and I also updated the address.

src/utils/cryptoassets.js
tima-t: Removed and amended in separate PR

monokh (approved):

PR #573, opened 2022-01-18
fix: reduce font size of token balance

Jennievon:

### DIFF
--- a/src/views/Account.vue
+++ b/src/views/Account.vue
@@ -24,7 +24,7 @@
             <span
               class="account-container_balance_value"
               :id="`${asset}_balance_value`"
-              :style="{ fontSize: formatFontSize(balance) }"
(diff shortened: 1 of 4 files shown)


src/views/Account.vue
```diff
@@ -24,7 +24,7 @@
             <span
               class="account-container_balance_value"
               :id="`${asset}_balance_value`"
-              :style="{ fontSize: formatFontSize(balance) }"
+              :style="{ fontSize: '30px' }"
```
monokh, line 27: Is there a reason this change isn't done in the css class?

monokh (changes requested):

src/views/Account.vue
Jennievon: This has been updated, @monokh.

monokh (approved):

PR #574, opened 2022-01-18
refactor: update wallet screens - onboarding

Jennievon:

[#LIQ-699](https://linear.app/liquality/issue/LIQ-699/integrate-tandc-and-success-screen-in-import-wallet-flow-missing)
[#LIQ-703](https://linear.app/liquality/issue/LIQ-703/align-pw-creation-step-in-wallet-creation-with-import-flow-and)


src/views/Splash.vue
```diff
@@ -14,7 +14,7 @@
             : { path: '/onboarding/home', query: { isImport: true } }
         "
         ><p class="text-center" id="import_with_seed_phrase_option">
-          Import with seed phrase
+          Forgot password? Import with seed phrase
```
monokh, line 17: If I just installed the wallet. It still say's "Forgot password?" But it shouldn't

src/views/UnlockWallet.vue
```diff
@@ -66,6 +66,7 @@ export default {
       password: ''
     }
   },
+  computed: mapState(['termsAcceptedAt']),
```
monokh, line 69: You added this but its not used?

monokh (changes requested):

src/views/Splash.vue
Jennievon: I had that implemented but the design shows this:

Screenshot 2022-01-18 at 15 43 21

src/views/UnlockWallet.vue
Jennievon: Taking it out now.

src/views/Splash.vue
monokh: i mean that's surely wrong? 😄  There is no password at that stage.

Jennievon: I agree. I have changed it to just "Import with seed phrase" now.

monokh (approved):

PR #575, opened 2022-01-18
feat: Enable eip1559 for matic mainnet

AndonMitev:

### DIFF
--- a/src/views/Send/Send.vue
+++ b/src/views/Send/Send.vue
@@ -209,7 +209,7 @@ import { mapState, mapActions, mapGetters } from 'vuex'
 import _ from 'lodash'
 import BN from 'bignumber.js'
 import cryptoassets from '@/utils/cryptoassets'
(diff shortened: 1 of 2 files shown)


Jennievon (approved):
LGTM

beemi (approved):

src/views/Send/Send.vue
```diff
@@ -381,10 +381,7 @@ export default {
       return BN(this.amount).plus(BN(this.currentFee))
     },
     isEIP1559Fees() {
-      return (
-        this.assetChain === 'ETH' ||
-        (this.assetChain === 'MATIC' && this.activeNetwork === 'testnet')
-      )
+      return this.assetChain === 'ETH' || this.assetChain === 'MATIC'
```
monokh, line 384: cryptoassets[this.asset].chain === ChainId.Ethereum is probably the correct way to do this
but it is the same result

AndonMitev: just replaced it

PR #577, opened 2022-01-18
fix: fee step on custom fee

tima-t:
# :books: liq-955 :books:


Jennievon:
@tima-t, please resolve conflict

PR #581, opened 2022-01-19
fix: Speed up ui changes

AndonMitev:
# :books: https://linear.app/liquality/issue/LIQ-961/the-speed-up-button-looks-like-a-text-than-a-button-when-a-send :books:

### DIFF
--- a/src/views/Details/TransactionDetails.vue
+++ b/src/views/Details/TransactionDetails.vue
@@ -56,14 +56,21 @@
                 >{{ assetChain }} Speed:
                 <span class="text-capitalize">{{ item.feeLabel }}</span></span
               >
-              <span>Fee: {{ item.fee }} {{ feeUnit }}</span>
+              <span>Fee: {{ itemFee }} {{ feeUnit }}</span>
               <span>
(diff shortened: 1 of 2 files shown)


src/views/Details/TransactionDetails.vue
```diff
@@ -160,7 +168,7 @@ export default {
     itemFee() {
       return typeof this.item.fee !== 'object'
         ? this.item.fee
-        : this.item.fee.suggestedBaseFeePerGas + this.item.fee.maxPriorityFeePerGas
+        : BN(this.item.fee.suggestedBaseFeePerGas + this.item.fee.maxPriorityFeePerGas).dp(3)
```
Jennievon, line 171: Hi @AndonMitev, Harsh shared this and mentioned the fee calculation is in a range and not specific.

image

AndonMitev: Hi @Jennievon this is not where we set the fee, this is what fee have been set by the user and it's used to be displayed. Lower and upper bounds are used when we manipulating the custom fees

Jennievon (approved):

monokh (approved):

PR #583, opened 2022-01-19
fix: liq 935 white screen error on add erc20

tima-t:
# :books liq-old-935 Ticket :books:


Jennievon (approved):

bradleySuira (approved):
LGTM

tima-t:
> It feels like we could end up missing this conversion somewhere as it's duplicated many times Is it not possible to do this in a central way? For example, can we instead just make sure symbols in cryptoassets are converted/don't have special symbols. Either inside the cryptoassets library or in the cryptoassets getter It might be possible also to use this as a middleware to do the conversion? https://router.vuejs.org/guide/advanced/navigation-guards.html

I agree - I am going to refactor it and only then we are going to merge in dev.

tima-t:
> It feels like we could end up missing this conversion somewhere as it's duplicated many times Is it not possible to do this in a central way? For example, can we instead just make sure symbols in cryptoassets are converted/don't have special symbols. Either inside the cryptoassets library or in the cryptoassets getter It might be possible also to use this as a middleware to do the conversion? https://router.vuejs.org/guide/advanced/navigation-guards.html

It is refactored according to your requirements. I think it is much better now. If it looks good to you - we can merge.

tima-t:
a new pr solving the problem is opened against the release candidate

PR #587, opened 2022-01-21
feat: Add eip1559 to be displayed when user is setting fee for dapp i…

AndonMitev:
feat: Add eip1559 to be displayed when user is setting fee for dapp interaction

# :books: [Linear Ticket](https://linear.app/liquality/issue/LIQ-970/gas-is-displaying-as-nan-and-network-fee-as-object-object-when-a) :books:

### DIFF
--- a/src/components/CustomFeesEIP1559.vue
+++ b/src/components/CustomFeesEIP1559.vue
@@ -3,7 +3,7 @@
     <NavBar :showBackButton="false" :showBack="false" :showMenuList="false">
       Network Speed/Fee
     </NavBar>
-    <div class="wrapper form" :class="{ wrapperPromiseSend: promiseSendView }">
+    <div class="wrapper form" :class="{ wrapperSend: padLabels }">
       <div class="wrapper_top">
         <div class="selected-asset" id="custom_fee_selected_asset">
           <img :src="getAssetIcon(asset)" class="asset-icon" />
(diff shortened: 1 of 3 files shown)


beemi (approved):

src/components/CustomFeesEIP1559.vue
```diff
@@ -3,7 +3,7 @@
     <NavBar :showBackButton="false" :showBack="false" :showMenuList="false">
       Network Speed/Fee
     </NavBar>
-    <div class="wrapper form">
+    <div class="wrapper form" :class="{ wrapperPromiseSend: promiseSendView }">
```
monokh, line 6: `promiseSendView`
that seems like a very weird name when what it does it add some padding. Can't you call it something more relevant? even sometihng like `padLabels`

I don't know the specifics but it doesn't seem appropriate for this component to look different based on where its embedded...

src/views/PermissionSend.vue
```diff
@@ -271,7 +289,12 @@ export default {
     },
     applyCustomFee({ fee }) {
       const presetFee = Object.entries(this.assetFees).find(
-        ([speed, speedFee]) => speed !== 'custom' && speedFee.fee === fee
+        ([speed, speedFee]) =>
+          speed !== 'custom' &&
```
monokh, line 293: These few lines are really hard to follow. anyway to improve it?

you can name statements for example

const feeMatches = speedFee.fee.maxPriorityFeePerGas === fee.maxPriorityFeePerGas

monokh (changes requested):

src/views/PermissionSend.vue
AndonMitev: Fixed, splited to 2 varialbes: `isEIP1559Fee` and `isLegacyFee` for better readability

src/components/CustomFeesEIP1559.vue
AndonMitev: applied mate

PR #606, opened 2022-01-24
fix: UI issues from QA report

AndonMitev:
# :books:[ Linear Ticket](https://linear.app/liquality/issue/LIQ-960/the-updown-arrow-while-setting-a-gas-on-network-fee-is-not-changing) :books:


beemi (approved):
LGTM

PR #607, opened 2022-01-24
test: LIQ-990

beemi:
test: 1inch v4 quote checks / fiat values check

LIQ-990


AndonMitev (commented):
LGTM

PR #614, opened 2022-01-26
fix: wallet flow

Jennievon:

### DIFF
--- a/src/views/Onboarding/OnboardingSetup.vue
+++ b/src/views/Onboarding/OnboardingSetup.vue
@@ -44,7 +44,7 @@
       v-show="currentStep === 'password'"
       :imported="imported"
       @on-unlock="onUnlock"
-      @currentStep="currentStep = imported ? $router.go(-1) : 'confirm'"
+      @currentStep="changeCurrentStep"
     />
     <Congratulations v-show="currentStep === 'congrats'" />
   </div>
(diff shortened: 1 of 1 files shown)


src/views/Onboarding/OnboardingSetup.vue
```diff
@@ -44,7 +44,7 @@
       v-show="currentStep === 'password'"
       :imported="imported"
       @on-unlock="onUnlock"
-      @currentStep="currentStep = 'confirm'"
+      @currentStep="currentStep = imported ? $router.go(-1) : 'confirm'"
```
monokh, line 47: something about this doesn't seem right. I can't understand what its doing because its setting and navigating at the same time
You should make this function separate and explicitly define the steps of behaviour such as 
```
if (currentStep === 'imported') router.go('/password')
else currentStep = 'confirm'
```

also try to avoid the $router.go(-1), usually its not very reliable. Route to a specific page you are looking to direct to

monokh (changes requested):

src/views/Onboarding/OnboardingSetup.vue
Jennievon: All fixed.

bradleySuira (approved):
LGTM

src/views/Onboarding/OnboardingSetup.vue
```diff
@@ -44,7 +44,7 @@
       v-show="currentStep === 'password'"
       :imported="imported"
       @on-unlock="onUnlock"
-      @currentStep="currentStep = 'confirm'"
+      @currentStep="changeCurrentStep"
```
monokh, line 47: I think this event should be renamed to "on-cancel"
here and in in OnboardingPassword

PR #619, opened 2022-01-27
refactor: password validation

Jennievon:

[#LIQ-703](https://linear.app/liquality/issue/LIQ-703/align-pw-creation-step-in-wallet-creation-with-import-flow-and)

### DIFF
--- a/src/views/Onboarding/OnboardingPassword.vue
+++ b/src/views/Onboarding/OnboardingPassword.vue
@@ -82,11 +82,7 @@ export default {
   computed: {
     passwordMatch() {
       if (!this.password || !this.confirmPassword) return false
-      if (
-        this.password.length === this.confirmPassword.length &&
-        this.password !== this.confirmPassword
-      )
-        return true
+      if (this.password !== this.confirmPassword) return true
       return false
     },
     disableNext() {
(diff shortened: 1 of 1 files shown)


src/views/Onboarding/OnboardingPassword.vue
```diff
@@ -82,11 +82,7 @@ export default {
   computed: {
     passwordMatch() {
       if (!this.password || !this.confirmPassword) return false
-      if (
-        this.password.length === this.confirmPassword.length &&
-        this.password !== this.confirmPassword
-      )
-        return true
+      if (this.password !== this.confirmPassword) return true
```
monokh, line 85: return true? If the passwords _dont_ equal. Shouldn't `passwordMatch` return false?

its also generally better to do things in this order:
```
if (...) return x
if (...) return x

return y
```
Where x or y can be true or false

monokh (changes requested):

src/views/Onboarding/OnboardingPassword.vue
Jennievon: Fixed.

monokh (approved):

PR #643, opened 2022-02-01
feat: Boost Provider ERC20 to Native asset

skeremidchiev:

[LIQ-1051](https://linear.app/liquality/issue/LIQ-1051/liquality-boost-provider-swaps-from-erc20-to-native)


src/build.config.js
```diff
@@ -58,9 +58,18 @@ export default {
         agent:
           process.env.VUE_APP_AGENT_TESTNET_URL || 'https://liquality.io/swap-testnet-dev/agent'
       },
-      liqualityBoost: {
-        name: 'Liquality Boost',
-        type: SwapProviderType.LIQUALITYBOOST,
+      // Native to ERC20
+      liqualityBoost_type1: {
```
tima-t, line 62: name is in mixed case "liqualityBoost_type1".
Should be in  camel case for consistency. 
Also I don't really like the name liqualityBoost_type1  and type2 - we should brainstorm/ ask UX team.

Last but not least keep in mind that changing the name of existing liquidity providers will need backup compatibility support.

src/swaps/liquality/LiqualitySwapProvider.js
```diff
@@ -183,7 +183,7 @@ class LiqualitySwapProvider extends SwapProvider {
 
   updateOrder(order) {
     return axios({
-      url: this.config.agent + '/api/swap/order/' + order.id,
+      url: this.config.agent + '/api/swap/order/' + order.orderId,
```
tima-t, line 186: Is this fixing of old bug? 
If we create a boost provider in the perfect case we should not change the liquality provider, just use it as it is.
If it is needed then it is fine, but be careful for regressions.

src/swaps/liqualityboost/lbspERC20toNative/LbspERC20toNative.js
```diff
@@ -0,0 +1,345 @@
...
+        swap: swapAutomatedMarketMakerFormat
+      })
+    }
+
+    if (!updates) return
+
+    return {
+      ...updates,
+      // reset from and to assets and values
+      from: swap.from,
```
tima-t, line 225: If you reset the from value isn't this resulting in showing wrong information in the timeline view. At the worst case we should keep the from value in some separate property I believe and handle it in the timeline view. 

E.g when you start a swap from sovryn -> eth , if you come to this step where you already exchanged sovryn to rbtc it will update the from value and in timeline diagram you will see that your swap is from rbtc -> eth, but user is swapping sov -> eth. I am not sure if I am correct but you should check this in the timeline view

src/swaps/liqualityboost/lbspERC20toNative/LbspERC20toNative.js
```diff
@@ -0,0 +1,345 @@
...
+      fromAmount: swap.fromAmount,
+      to: swap.to,
+      // keep `toAmount` (from updates object) only in case swap transitioned from AMM to LSP
+      toAmount: updates.status === 'INITIATED' ? updates.toAmount : swap.toAmount
```
tima-t, line 229: same as the above comment, regarding the returned properties as they are also use in timeline diagram if I am not mistaken

src/swaps/liquality/LiqualitySwapProvider.js
skeremidchiev: It's not a bug. It's the only way to make things work with minimal changes (that came to my mind).
`orderId` and `id` are the same thing. But in case of swapping from erc20 to native we need to keep the same `id` between bot swaps (the one over AMM and over LSP). Otherwise wallet cant pull the correct swap data from storage.

src/swaps/liqualityboost/lbspERC20toNative/LbspERC20toNative.js
skeremidchiev: No. It's quite the opposite. All 'initial' data is kept in 'swap' objet completely intact. In the start of `performNextSwapAction` method the two objects `swapLiqualityFormat` and  `swapAutomatedMarketMakerFormat` format the data so it will be suitable for the AMM or LSP. At the end of the method data is again formatted in a way that could be correctly shown in time line.

src/swaps/liqualityboost/lbspERC20toNative/LbspERC20toNative.js
skeremidchiev: This a specific case in which swap transitions from AMM to LSP. In this case due to slippage the `toAmount` seems to go a little bit lower. That's why the new value is kept.

src/build.config.js
skeremidchiev: Fixed. Liquality boost in included in `mapLegacyProvidersToSupported`.

monokh (changes requested):

src/store/getters.js
```diff
@@ -45,7 +45,8 @@ const TESTNET_ASSETS = [
 }, {})
 
 const mapLegacyProvidersToSupported = {
-  oneinchV3: 'oneinchV4'
+  oneinchV3: 'oneinchV4',
```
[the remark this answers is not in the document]
skeremidchiev: This is simply mapping older versions to the new one.

PR #665, opened 2022-02-10
fix: format-send-amount-on-permission-send-view

tima-t:
# :books:   LIQ-old-1068  :books:


tima-t:
PR closed in favour of new one against the current release candidate

PR #703, opened 2022-03-04
feat: Add fallback if env is missing for terra

AndonMitev:
feat: Add fallback if env is missing for terra


beemi:
LGTM, please create pr against latest releases candidate

beemi:
This is done in another push https://github.com/liquality/wallet/commit/879a22d8288185c54047ccdda1caaffac3d799f6

PR #706, opened 2022-03-08
feat: NEAR swap ma amount

skeremidchiev:

[LIQ-607](https://linear.app/liquality/issue/LIQ-607/max-near-swaps-seem-broken-as-of-prod-v048)

### DIFF
--- a/src/swaps/liquality/LiqualitySwapProvider.js
+++ b/src/swaps/liquality/LiqualitySwapProvider.js
@@ -164,6 +164,19 @@ class LiqualitySwapProvider extends SwapProvider {
       return mapValues(totalFees, (f) => unitToCurrency(cryptoassets[asset], f))
     }
 
+    if (txType === LiqualitySwapProvider.txTypes.SWAP_INITIATION && asset === 'NEAR') {
+      const fees = {}
+      // default storage fee recommended by NEAR dev team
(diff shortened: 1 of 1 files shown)


monokh (approved):

src/swaps/liquality/LiqualitySwapProvider.js
```diff
@@ -164,6 +164,19 @@ class LiqualitySwapProvider extends SwapProvider {
       return mapValues(totalFees, (f) => unitToCurrency(cryptoassets[asset], f))
     }
 
+    if (txType === LiqualitySwapProvider.txTypes.SWAP_INITIATION && asset === 'NEAR') {
+      const fees = {}
+      // default storage fee recommended by NEAR dev team
+      // It leaves 0.02$ dust in the wallet on max value
+      const storageFee = BN(0.00125)
+      for (const feePrice of feePrices) {
+        fees[feePrice] = getTxFee(LiqualitySwapProvider.feeUnits[txType], asset, feePrice).plus(
```
monokh, line 173: Shouldn't it also check that `max` is `true`?

AndonMitev: Nah, because we need to pay for this storage every time, just before when Max were not selected users were left with some amount of funds and they were able to cover the storage fee

PR #712, opened 2022-03-16
Feature remove ledger bridge

bradleySuira:


src/router/index.js
```diff
@@ -248,11 +284,12 @@ const router = new VueRouter({
...
 
-const waitForBrokerReady = async (_to, _from, next) => {
+router.beforeEach(async (to, from, next) => {
+  // wait for the broker
   await broker.ready.promise
-  next()
-}
-
-router.beforeEach(waitForBrokerReady)
+  if (!store.state.unlockedAt && to.meta?.protect == true && to.name !== 'Splash') {
```
monokh, line 290: This should handle any route without `meta: { protect }` as `protect = true`. Incase we forget to add protect to a route

src/store/modules/app/executeRequest.js
```diff
@@ -41,30 +39,32 @@ export const executeRequest = async ({ rootGetters, dispatch, rootState }, { req
...
-        })
-      } else {
-        resolve(call)
-      }
-    } else {
-      resolve(call)
-    }
+    resolve(call)
+
+    // const { ledgerBridgeConnected, ledgerBridgeTransportConnected } = rootState.app
```
monokh, line 44: What about injection though? Does ledger integration work with injection after this change?

src/views/Swap/Swap.vue
```diff
@@ -502,55 +500,102 @@ export default {
...
+        provider,
+        userSelectedQuote
+      } = this.$route.query
+      this.currentStep = currentStep
+      this.maxOptionActive = maxOptionActive
+      this.toAccountId = toAccountId
+      _toAsset = toAsset
+      _sendAmount = sendAmount
+
+      this.selectedFee = selectedFees.split(',').reduce((prev, curr) => {
```
monokh, line 534: instead of splitting `,` and `=`
it might be better to parse it first using `qs` library, then do the rest

monokh (changes requested):

src/utils/hardware-wallet/utils.js
```diff
@@ -0,0 +1,83 @@
+import TransportWebHID from '@ledgerhq/hw-transport-webhid'
+import { Debug } from '@liquality/debug'
+import { WalletError } from '@liquality/errors'
+const debug = Debug('ledger-transport')
+
+let transport = null
+let onDisconnectCallbacks = []
+export const LEDGER_USB_VENDOR_ID = '0x2c97'
```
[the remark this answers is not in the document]
bradleySuira: yes this apply for all ledger at this moment I tested with nano and nano + but should work wit others

src/router/index.js
```diff
@@ -248,11 +284,12 @@ const router = new VueRouter({
...
+    next({ name: 'Splash' })
```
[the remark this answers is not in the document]
bradleySuira: I set it because currently is the page that we redirect, with the unlock/login or Create wallet options

src/router/index.js
bradleySuira: yes if we forget to add the meta field should be ok `to.meta?.protect == true`

src/utils/hardware-wallet/utils.js
```diff
@@ -0,0 +1,83 @@
+import TransportWebHID from '@ledgerhq/hw-transport-webhid'
```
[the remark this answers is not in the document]
bradleySuira: this was moved to wallet-core + chaining

src/store/modules/app/executeRequest.js
bradleySuira: right it should work, the connection request should be in the popup and after that we can call the client directly

src/views/Swap/Swap.vue
bradleySuira: fixed

src/views/Send/Send.vue
```diff
@@ -605,16 +613,40 @@ export default {
...
+      }
+      if (currentStep) {
+        this.currentStep = currentStep
+      }
+      if (maxOptionActive) {
+        this.maxOptionActive = maxOptionActive === 'true' ? true : false
+      }
+    }
+    this.updatingFees = true
+    debugger
```
monokh, line 641: debugger :S
Is this code ok?

src/views/Swap/Swap.vue
```diff
@@ -1251,20 +1267,38 @@ export default {
...
       this.quotes = []
       this.updatingQuotes = true
-      this._updateQuotes()
+      this.debounceUpdateQuotes()
     },
-    selectQuote(provider) {
+    setQuoteProvider(provider) {
       const matchingQuote = this.quotes.find((q) => q.provider === provider)
       this.selectedQuote = matchingQuote
+      console.log('this.selectedQuote', this.matchingQuote)
```
monokh, line 1281: console log?

monokh (approved):
LGTM 
but there is some `console.log` and `debugger`

src/views/Send/Send.vue
bradleySuira: fixed :)

src/views/Swap/Swap.vue
bradleySuira: fixed

PR #725, opened 2022-03-24
feat: prevent native assets disabling

skeremidchiev:

[LIQ1192](https://linear.app/liquality/issue/LIQ-1192/disabling-btc-will-break-swap-feature)


kraikov:
I think we were supposed to fix that, instead of preventing disabling?

PR #728, opened 2022-03-28
fix: correct deafult asset logic

skeremidchiev:

[LIQ-1192](https://linear.app/liquality/issue/LIQ-1192/disabling-btc-will-break-swap-feature)

### DIFF
--- a/src/views/Swap/Swap.vue
+++ b/src/views/Swap/Swap.vue
@@ -489,20 +489,25 @@ export default {
       await this.updateMaxSwapFees()
     })()
 
-    // Try to use the same account for (from and to) if it has more than one asset
     let toAsset = null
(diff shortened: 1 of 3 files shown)


src/views/Swap/Swap.vue
```diff
@@ -489,20 +489,25 @@ export default {
...
       toAsset = this.account?.assets.find((a) => a !== this.asset)
     } else {
-      if (this.networkAccounts.length > 0) {
-        const toAccount = this.networkAccounts.find(
+      // use another account
+      if (this.accountsData.length > 0) {
+        const toAccount = this.accountsData.find(
           (account) =>
-            account.assets && !account.assets.includes(this.asset) && account.id !== this.accountId
+            account.assets &&
```
monokh, line 503: These statements could really use some naming 😄  hard to tell what it is doing

monokh (approved):

PR #731, opened 2022-03-31
fix: Gas limit for uniswap

AndonMitev:
# :books: [Linear Ticket](https://linear.app/liquality/issue/LIQ-1034/getting-[object-object]-gwei-error-on-custom-fees-and-when-user-click) :books:


AndonMitev:
Part of wallet core, no more needed here

PR #735, opened 2022-04-02
feat: componentize copy address tooltip

Jennievon:

[#LIQ-704](https://linear.app/liquality/issue/LIQ-704/update-all-shortened-account-addresses-with-[6-digits][4digits]-and)

Screenshot 2022-04-03 at 02 16 17
Screenshot 2022-04-03 at 02 17 04


bradleySuira (changes requested):
Hi @Jennievon all GTM but I think that you should use RC 0.59.0 branch as a target branch, cc: @beemi

PR #740, opened 2022-04-05
fix: Multiple account on terra

AndonMitev:
# :books: [Linear Ticket](https://linear.app/liquality/issue/LIQ-647/multi-account-when-adding-the-multiple-accounts-for-terra-same-account) :books:


bradleySuira (approved):
LGTM

skeremidchiev (approved):

PR #758, opened 2022-04-22
fix: improved logic  for amm validation

tima-t:
# :books:LIQ-1272 :books:

### DIFF
--- a/src/views/Swap/Swap.vue
+++ b/src/views/Swap/Swap.vue
@@ -725,12 +725,11 @@ export default {
 
       const account = isERC20(this.asset) ? this.account : this.toAccount
       const balance = account?.balances[this.selectedQuote.bridgeAsset]
-      const toSwapFeeInUnits = currencyToUnit(
+      const SwapFeeInUnits = currencyToUnit(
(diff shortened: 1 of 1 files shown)


src/views/Swap/Swap.vue
```diff
@@ -725,12 +725,11 @@ export default {
...
       const balance = account?.balances[this.selectedQuote.bridgeAsset]
-      const toSwapFeeInUnits = currencyToUnit(
+      const SwapFeeInUnits = currencyToUnit(
         cryptoassets[this.selectedQuote.bridgeAsset],
-        this.receiveFee
+        isERC20(this.asset) ? this.fromSwapFee : this.receiveFee
       )
-
-      return BN(balance).gt(toSwapFeeInUnits)
+      return BN(balance).gt(SwapFeeInUnits)
```
monokh, line 732: Why is this not camel case anymore?

monokh (approved):

src/views/Swap/Swap.vue
tima-t: should be with camelCase - correct 
I will fix it in another PR as this was merged already...

PR #760, opened 2022-04-22
streamline onboarding experience

Jennievon:

[#LIQ-768](https://linear.app/liquality/issue/LIQ-768/streamline-onboarding-experience)

### DIFF
--- a/src/components/AnalyticsOptInModal.vue
+++ b/src/components/AnalyticsOptInModal.vue
@@ -20,7 +20,7 @@
             name="optin_anaylitics_option"
             id="optin_anaylitics_accept"
             autocomplete="off"
-            :checked="analyticsAccepted === true"
+            :checked="accepted === true"
             @click="accept"
             :disabled="loading"
           />
@@ -35,7 +35,7 @@
             name="optin_anaylitics_option"
             id="optin_anaylitics_reject"
             autocomplete="off"
-            :checked="analyticsAccepted === false"
+            :checked="accepted === false"
             @click="reject"
             :disabled="loading"
           />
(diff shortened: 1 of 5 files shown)


monokh (approved):

src/components/AnalyticsOptInModal.vue
```diff
@@ -23,14 +59,14 @@
...
-
 export default {
   components: {
     Modal
   },
   data: function () {
     return {
-      loading: false
+      loading: false,
+      analyticsAccepted: null
```
bradleySuira, line 69: In this case you can access to `state.analytics?.acceptedDate` and use `mappState` to access it as a computed property

src/components/AnalyticsOptInModal.vue
```diff
@@ -46,19 +82,20 @@ export default {
     ...mapActions(['setAnalyticsResponse', 'initializeAnalytics']),
     close() {
       this.setAnalyticsOptInModalOpen({ open: false })
+      this.$emit('goToSetup')
     },
     async accept() {
       this.loading = true
+      this.analyticsAccepted = true
```
bradleySuira, line 89: Not need it if you use the acceptedDate from the state

bradleySuira (changes requested):

skeremidchiev:
What is the correct behaviour of the close (x) button?

Screenshot 2022-04-26 at 13 39 25

It seems like it brings you to next step.

Jennievon:
> What is the correct behaviour of the close (x) button?
> 
> Screenshot 2022-04-26 at 13 39 25
> 
> It seems like it brings you to next step.

Fixed now.

src/components/AnalyticsOptInModal.vue
Jennievon: > Not need it if you use the acceptedDate from the state

Fixed but on `reject`, the state value is `null`. Have to add a data property for accepting. Pushed a fix. Kindly review.

Jennievon:
@skeremidchiev @bradleySuira, please review

PR #772, opened 2022-05-09
fix: Fix update balance amplitude event

beemi:

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -136,7 +136,7 @@ store.subscribe(async ({ type, payload }, state) => {
         properties: {
           category: 'Dapps',
           action: 'Dapp Injected',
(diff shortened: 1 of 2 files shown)


src/background.js
```diff
@@ -5,16 +5,40 @@ import { wait } from './store/utils'
...
 
 function asyncLoop(fn, delay) {
   return wait(delay())
     .then(() => fn())
     .then(() => asyncLoop(fn, delay))
 }
 
+function getBalance(state) {
+  let total = 0
+  state.accounts?.[state.activeWalletId]?.[state.activeNetwork].map((item) => {
```
monokh, line 20: I think this is basically `reduce` :D

monokh (approved):

PR #773, opened 2022-05-10
fix: correct min amount for boost

skeremidchiev:

[LIQ-1291](https://linear.app/liquality/issue/LIQ-1291/no-quote-is-appearing-for-bpro-to-any-other-rsk-token-as-the-min-of)

### DIFF
--- a/src/views/Swap/Swap.vue
+++ b/src/views/Swap/Swap.vue
@@ -657,8 +657,6 @@ export default {
       return !!liqualityMarket
     },
     min() {
-      if (!this.fiatRates[this.asset]) return BN(0)
-
       const toQuoteAsset =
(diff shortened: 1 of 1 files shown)


src/views/Swap/Swap.vue
```diff
@@ -688,6 +686,8 @@ export default {
       } else {
         min = liqualityMarket
           ? BN(liqualityMarket.min)
+          : !this.fiatRates[this.asset]
```
tima-t, line 689: I am not a big fan of these nested ternary operators (even if it is shorter by code perspective).
I will approve it, but we should consider refactoring this later

tima-t (approved):

PR #774, opened 2022-05-10
feat: add support for signing typed data via injection

kraikov:

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -23,7 +23,7 @@ class ProviderManager {
   proxy (type, data) {
     return new Promise((resolve, reject) => {
       const id = Date.now() + '.' + Math.random()
-  
+
       window.addEventListener(id, ({ detail }) => {
         const response = JSON.parse(detail)
         if (response.error) reject(new Error(response.error))
@@ -32,7 +32,7 @@ class ProviderManager {
         once: true,
         passive: true
       })
-  
+
       window.postMessage({
         id,
         type,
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -99,6 +107,7 @@ async function handleRequest (req) {
 
 window[injectionName] = {
   isLiquality: true,
+  isMetaMask: true, // OpenSea wants this
```
monokh, line 110: If we add this, we are going to be treated as metamask in all applications, and the liquality logo will no longer display. I think we should enquire with @alokm on this as i'm not sure this will be acceptable

src/inject.js
```diff
@@ -79,10 +79,18 @@ async function handleRequest (req) {
...
-  if(req.method === 'personal_sign') { 
+  if(req.method === 'personal_sign') {
     const sig = await eth.getMethod('wallet.signMessage')(req.params[0], req.params[1])
     return '0x' + sig
   }
+
+  if(req.method === 'eth_signTypedData' ||
+    req.method === 'eth_signTypedData_v3' ||
+    req.method === 'eth_signTypedData_v4') {
+    const sig = await eth.getMethod('wallet.signTypedMessage')(req)
```
monokh, line 90: Shouldn't this be added to the list of `CONFIRM_REQUIRED` requests???
I also see no UI changes, but this type of action shouldn't be automatically signed by the wallet and should have a confirm prompt.

monokh (changes requested):

src/inject.js
kraikov: Without it whenever the "Metamask" button in Opensea is clicked, the user is redirected to https://metamask.io/download/

skeremidchiev (approved):
LGTM

AndonMitev (approved):

PR #775, opened 2022-05-10
feat: integrate hop exchange in wallet

tima-t:
# :books: liq-1100 :books:


beemi:
@tima-t  will this PR takes wallet-core 1.5.0 as well?

PR #786, opened 2022-05-24
fix: temporary solution to ensure transaction fee is covered when max…

tima-t:
temporary solution to ensure transaction fee is covered when max

# :books: LIQ-1260 :books:


monokh (approved):
looks ok , but really need to address this properly

PR #795, opened 2022-06-01
feat: improve minimum amount logic for boosts

skeremidchiev:
[LIQ-1393](https://linear.app/liquality/issue/LIQ-1393/liquality-boost-or-reduce-the-min-for-pweth-to-btc-so-that-user-can)
[LIQ-1062](https://linear.app/liquality/issue/LIQ-1062/user-need-to-increase-the-input-amount-to-see-the-liquidity-between)


AndonMitev (approved):

bradleySuira (commented):
LGTM

PR #808, opened 2022-06-14
fix: get rid of duplicate balance update dispatching

kraikov:
the `updateBalances` action is dispatched twice during wallet unlock:

1. once during `Wallet.vue` component creation (in `created()`)
2. once in the `UNLOCK_WALLET` listener in `background.js`

When the wallet is unlocked the first component is always `Wallet` so we can get rid of the dispatch that happens in `UNLOCK_WALLET`


monokh (approved):

monokh:
Tests reporting balance issue. Might be related to your change?

PR #813, opened 2022-06-23
Liq 1205/error messaging chainify

tima-t:

### DIFF
--- a/src/components/NoFundsForNetworkFee.vue
+++ b/src/components/NoFundsForNetworkFee.vue
@@ -1,0 +1,64 @@
+<template>
+  <div class="notification-content">
+    <div v-if="!receiveFeeProblem && isSameChain" class="notification-text">
+      Not enough {{ assetChain }} to cover fees. Lower speed to account for gas or get more.
+    </div>
+    <div v-if="receiveFeeProblem || !isSameChain" class="notification-text">
+      Lower speed to account for gas to cover both sides of transaction, or get more.
+    </div>
+    <br />
+    <div v-if="isSameChain || receiveFeeProblem">
+      <router-link :to="accountUrl" class="btn btn-option get-bridgeAsset-btn">
+        Get {{ assetChain }}
+      </router-link>
+    </div>
+    <div v-else-if="!isSameChain">
+      <router-link :to="accountUrl" class="btn btn-option get-bridgeAsset-btn">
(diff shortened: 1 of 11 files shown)


src/components/NoFundsForNetworkFee.vue
```diff
@@ -0,0 +1,63 @@
...
+.notification-content {
+  // display: flex;
+  justify-content: space-between;
+  align-items: center;
+  .get-bridgeAsset-btn {
+    max-width: 120px;
+    padding: 0 0.5rem;
+    display: inline-block;
+  }
+  .notification-text {
```
AndonMitev, line 59: i think we can remove this block with styles

src/components/NoFundsForNetworkFee.vue
```diff
@@ -0,0 +1,63 @@
...
+    isSameChain() {
+      return this.assetChain === this.toAssetChain
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+.notification-content {
+  // display: flex;
```
AndonMitev, line 51: remove commented

src/views/Swap/SendInput.vue
```diff
@@ -69,7 +69,8 @@
           <v-popover offset="1" trigger="hover focus" class="mr-2">
             <button
               :class="{
-                active: amountOption === 'min' && !disabled
+                active: amountOption === 'min' && !disabled,
+                isNotVisible: min <= 0
```
AndonMitev, line 73: maybe it's easier to do it with _conditional rendering_ but if when "Min" button is hidden, position of Max is changed then this is fine as well

AndonMitev (approved):

src/views/Swap/SendInput.vue
tima-t: Yes, position of max is changed, if min is hidden

src/components/NoFundsForNetworkFee.vue
tima-t: Maybe I can remove display:inline-block, but the padding I need, because there was a UI issue linked to that

src/components/NoFundsForNetworkFee.vue
tima-t: done

PR #823, opened 2022-06-28
feat: addd priv-key-precommit

beemi:


kraikov (approved):

kraikov:
can we lock the version?

kraikov:
can we lock the version?

kraikov (changes requested):
can we lock the version?

beemi:
> can we lock the version?

its done

kraikov (approved):

PR #825, opened 2022-06-29
fix: Display cannot cover minimum message condition

AndonMitev:
# :books: [Linear Ticket](https://linear.app/liquality/issue/LIQ-1492/in-case-user-balance-is-less-than-min-balance-during-swap-then-correct) :books:

# :books: [Linear Ticket](https://linear.app/liquality/issue/LIQ-1489/gettoken-button-and-message-isnt-displayed-even-when-user-has) :books:

### DIFF
--- a/src/views/Swap/Swap.vue
+++ b/src/views/Swap/Swap.vue
@@ -65,7 +65,7 @@
           :asset="selectedQuote.bridgeAsset"
         />
       </InfoNotification>
-      <InfoNotification v-else-if="cannotCoverMinimum && insufficientFundsError">
+      <InfoNotification v-else-if="cannotCoverMinimum || insufficientFundsError">
         <CannotCoverMinimumMessage :asset="asset" :account-id="account.id" />
       </InfoNotification>
       <InfoNotification v-else-if="ethRequired && !insufficientFundsError">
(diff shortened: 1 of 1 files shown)


src/views/Swap/Swap.vue
```diff
@@ -65,7 +65,7 @@
           :asset="selectedQuote.bridgeAsset"
         />
       </InfoNotification>
-      <InfoNotification v-else-if="cannotCoverMinimum && !insufficientFundsError">
+      <InfoNotification v-else-if="cannotCoverMinimum || insufficientFundsError">
```
tima-t, line 68: Not sure if the condition should be changed like this. 

Current logic was  - if you  Cannot cover minimum && there is no Inssuficient Funds error - then show some message 

Now it is  - if you cannot cover minimum or   Inssuficient Funds error - it means that if the funds are inssuficient it will show an error similar to the error when use r cannot cover minimum.  With our new implementation when insufficientFundsError is triggered - the review button in the bottom left stays disabled in status insufficient funds and it does not  result in showing extra error message

src/views/Swap/Swap.vue
```diff
@@ -1074,6 +1077,17 @@ export default {
       this.fromAssetChanged(toAccountId, toAsset, toAmount)
       this.toAssetChanged(fromAccountId, fromAsset)
     },
+    async _getLiqualityLiquidityForAsset() {
```
tima-t, line 1080: One tricky case might be : 

I am trying to use liquality boost to swap BTC for PUSDC 

It will swap BTC for Matic, then matic for pusdc in 1 inch. 

In current check it won't be a problem because we are checking if the swap proider is 'liquality', but maybe we should check if the  liquality provider is NativeToERC20LiqualityBoost, but then the asset: this.toAsset, should be bridge asset or nativeAsset[ this.toAsset] instead of this.toAsset

src/views/Swap/Swap.vue
AndonMitev: Yeah, the requirements from the linear task are exactly this, expecting to display error if there are insufficient funds::

https://linear.app/liquality/issue/LIQ-1492/in-case-user-balance-is-less-than-min-balance-during-swap-then-correct

src/views/Swap/Swap.vue
```diff
@@ -867,10 +872,7 @@ export default {
     },
     insufficientFundsError() {
       const amount = BN(this.safeAmount)
-      if (amount.gt(this.max)) {
-        return true
-      }
-      return false
+      return amount.gt(this.max) || amount.lt(this.min)
```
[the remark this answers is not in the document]
AndonMitev: checking this one

PR #837, opened 2022-07-22
fix: added render improvements

bradleySuira:


src/views/Open.vue
```diff
@@ -11,6 +11,7 @@ export default {
   },
   methods: {
     unlocked() {
+      console.log('unlocked', new Date())
```
monokh, line 14: log

bradleySuira: fixed

PR #844, opened 2022-07-28
feat: unstoppable domain integration

amritkumarj:
image
image


src/utils/uns.js
```diff
@@ -0,0 +1,95 @@
+import { ChainId } from '@liquality/cryptoassets'
+const { default: Resolution } = require('@unstoppabledomains/resolution')
+
+const resolutionService = 'https://unstoppabledomains.g.alchemy.com/domains/'
+const tldAPI = 'https://resolve.unstoppabledomains.com/supported_tlds'
+const alchemyKey = 'bKmEKAC4HJUEDNlnoYITvXYuhrIshFsa'
```
monokh, line 6: When in `wallet-core` 
this should be in `build.config`

src/utils/uns.js
```diff
@@ -0,0 +1,95 @@
...
+
+function preparedDomain(domain) {
+  const retVal = domain ? domain.trim().toLowerCase() : ''
+  if (!reg.test(retVal)) {
+    throw 'Invalid domain name'
+  }
+  return retVal
+}
+
+export async function reverseUNS(address) {
```
monokh, line 88: Once this is over in `wallet-core` it would be good if this class inherited a standard interface. 

e.g. 
```
interface NameResolver {

public reverseLookup(address): {}
public lookupDomain(domain): {}
public isValidDomain(domain): {}
```
this way, the UI will not be tied to UNS, and we can implement other name resolvers following this pattern

src/views/Send/Send.vue
```diff
@@ -378,7 +400,14 @@ export default {
       return unit
     },
     isValidAddress() {
-      return chains[cryptoassets[this.asset].chain].isValidAddress(this.address, this.activeNetwork)
+      const isValidUNS = this.isValidUNSDomain
+      if (!isValidUNS) {
+        this.getUNSAddress()
```
monokh, line 405: As before, state changing function cannot be in computed

src/views/Send/Send.vue
```diff
@@ -192,7 +192,13 @@
           <div class="mt-40">
             <label>Send To</label>
             <p class="confirm-address" id="confirm-address">
-              {{ this.address ? shortenAddress(this.address) : '' }}
+              {{
```
monokh, line 195: When the expression is this big, it should be in a computed property

monokh (changes requested):

amritkumarj:
new pr #854

PR #854, opened 2022-08-08
feat: unstoppable domain integration 

amritkumarj:
First, need wallet-core pr merged https://github.com/liquality/wallet-core/pull/110

image
image


src/views/Send/Send.vue
```diff
@@ -36,6 +36,7 @@
                 placeholder="Address"
                 autocomplete="off"
                 required
+                @input="getDomainAddress"
```
monokh, line 39: Can you add `debounce` to this? It makes a network call and it would be slow an expensive to call it for every character the user edits in the field

src/views/Send/Send.vue
```diff
@@ -378,7 +391,10 @@ export default {
       return unit
     },
     isValidAddress() {
-      return chains[cryptoassets[this.asset].chain].isValidAddress(this.address, this.activeNetwork)
+      return (
+        this.isValidDomain ||
+        chains[cryptoassets[this.asset].chain].isValidAddress(this.address, this.activeNetwork)
+      )
     },
     addressError() {
```
monokh, line 399: When I enter a domain that can't be resolved (doesn't exist), I get error:
`'Wrong format. Please check the address.'`
But in this case it should say `Name not found` or something along those lines

monokh (changes requested):
Can you also update the wallet-core version. It is updated

src/views/Send/Send.vue
amritkumarj: It won't call network on every character edit, it only call if the domain has a valid TLD

src/views/Send/Send.vue
amritkumarj: For this, we need to call isValidTld function to check if this is a domain or gibberish. Which is an async function and would add unwanted complexity

src/views/Send/Send.vue
monokh: Even for valid TLD it shouldn't be calling per change, debounce is a good practice for input calling some kind of async. In the future we cannot assume what this function will be doing under the hood. For example if implementation changes, or other name resolver.

monokh:
If you update wallet-core version , we can start testing this.

monokh (approved):

monokh:
One question @amritkumarj 
For arbitrum, the ethereum address is resolved, is that expected behaviour?

monokh:
Another issue is that polygon addresses are not resolving:
Try https://ud.me/blockdudes.blockchain or https://ud.me/0BTCLIFE.NFT

PR #857, opened 2022-08-11
feat: Solana-integration

AndonMitev:
Solana Integration


kraikov (approved):

bradleySuira (approved):
LGTM, just Think that can be good to fix the build settings for wallet-core to use the imports in another way:  instead of `@liquality/wallet-core/dist/src/utils/address ` can be like `@liquality/wallet-core/utils/address`

Jennievon (approved):

PR #862, opened 2022-08-31
Feature/ext 31/localization support

bradleySuira:
# :books: Linear Ticket :books: #ext-31
This PR includes:
- Localization support 
- Base English and Spanish keys
- Localization switch inside settings page


src/locales/en/common.json
```diff
@@ -0,0 +1,90 @@
...
+    "type": "Type",
+    "continue": "Continue",
+    "next": "Next",
+    "back": "Back",
+    "new": "New",
+    "status": "Status",
+    "start": "Start",
+    "end": "End",
+    "completed": "Completed",
+    "advanved": "Advanved",
```
stephien, line 12: Hey @bradleySuira , both the code and word in English have a typo. We’ll fix these two typos in the common.json file using Crowdin, but I think you need to fix it here and other places.

src/locales/en/pages.json
```diff
@@ -0,0 +1,233 @@
...
+        "assets": "Assets",
+        "activity": "Activity",
+        "nfts": {
+            "zero": "NFTs (0)",
+            "one": "NFTs",
+            "other": "NFTs (%{count})"
+        },
+        "nftActivityDescription": "Once you start using your NFTs you will see the activity here",
+        "nftDescription": "Once you start owning NFTs with accounts in your Liquality wallet you will see them here",
+        "exploreNFT": "Explore NFTs on Polygon, Ethereum and Solana",
```
stephien, line 111: Shouldn’t Arbitrum also be mentioned here?

bradleySuira: thanks @stephien in this case I just copy the actual content from the wallet texts but if we don't need to mention Arbitrum I think that I can remove it just let me know.

src/locales/en/common.json
bradleySuira: thanks @stephien fixed now in all files

vue.config.js
```diff
@@ -82,9 +88,6 @@ module.exports = {
       manifestTransformer: (manifest) => {
         manifest.content_security_policy =
           "script-src 'self' 'unsafe-eval' 'unsafe-inline' https://cdn.segment.com 'sha256-ZgDy59Dh4jH9g/vcPYFpoQ1wumB4IdPEOS1BJc08i+Y='; object-src 'self';"
-        manifest.externally_connectable = {
```
monokh, line 85: why was it removed? it won't affect anytrhing?

monokh (approved):
LGTM

vue.config.js
bradleySuira: yes in this case we are not using anymore the ledger bridge so we don't need this settings

PR #870, opened 2022-09-05
NFT Updates

Jennievon:
- remove NFT from the swap asset list
- make the top bar fixed on the scroll

[LIQ-1556](https://linear.app/liquality/issue/LIQ-1556/title-bar-and-buttons-on-the-bottom-should-be-sticky-on-nft-screens)
[LIQ-1553](https://linear.app/liquality/issue/LIQ-1553/nft-option-should-not-display-under-select-asset-for-swap)


bradleySuira (approved):
LGTM

PR #874, opened 2022-09-15
feat: easy evm integrations

kraikov:
What remains to be done, but it cannot be done at the moment is the asset icons management:
https://github.com/liquality/wallet/blob/release-candidate-0.73.0/src/utils/accounts.js

I believe the UI/UX team is supposed to provide icons for:
- chain
- account

and we need an unified way to get the icon.

The ultimate result must be something like:

```ts
export const getAccountIcon = (chainId) => {
  return getAssetIcon(`${chainId}_account`)
}
export const getChainIcon = (chainId) => {
  return getAssetIcon(`${chainId}_chain`)
}
```


bradleySuira (approved):
LGTM

Jennievon (approved):
LGTM!

PR #879, opened 2022-09-22
fix: fixed ledger to ledger send for custom fees and blank screen error

bradleySuira:
Fixes for #EXT-78 and #EXT-79 for ledger screens


Jennievon (approved):
LGTM!

src/views/Send/Send.vue
```diff
@@ -642,19 +641,24 @@ export default {
       const { amount, address, selectedFee, currentStep, maxOptionActive, customFee } = sendParams
       this.amount = amount
       this.address = address
+      if (maxOptionActive) {
+        this.maxOptionActive = maxOptionActive === 'true' ? true : false
```
Jennievon, line 645: @bradleySuira why aren't we doing `this.maxOptionActive = maxOptionActive` here though?

bradleySuira: it's because the library that we are using `qs` when parse the data back convert all values to string, it's really handy but have that issue

PR #897, opened 2022-10-12
fix: assets must be fetched from store instead of cryptoassets

kraikov:

### DIFF
--- a/src/broker/Background.js
+++ b/src/broker/Background.js
@@ -188,7 +188,7 @@ class Background {
     let setDefaultEthereum = false
     let { chain, asset } = data
     if (asset) {
-      chain = getAsset(activeNetwork, asset).chain
+      chain = this.store.getters.cryptoassets(activeNetwork, asset).chain
     }
     if (!chain) {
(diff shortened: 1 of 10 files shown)


src/broker/Background.js
```diff
@@ -188,7 +188,7 @@ class Background {
     let setDefaultEthereum = false
     let { chain, asset } = data
     if (asset) {
-      chain = getAsset(activeNetwork, asset).chain
+      chain = this.store.getters.cryptoassets(activeNetwork, asset).chain
```
bradleySuira, line 191: maybe can be `this.store.getters.cryptoassets(activeNetwork, asset)?.chain`

bradleySuira (approved):
LGTM, just please check the cases when maybe we don't have the asset inside crypto asset so the chain id will be null or undefined

src/broker/Background.js
kraikov: I agree. However there shouldn't be a case in which an asset is used and it's missing in the store. If that happens, there is a bigger, greater issue that must be fixed.

monokh (approved):

PR #901, opened 2022-10-17
fix: support liquality errors

koderholic:
This PR updates does the following 

1. Backward compatibility for errors:  Some Errors returned from wallet core are `instances of` Liquality Error(extension of Js Error), others are `instances of` Js Error.  So `background.js,`  `replyPermission.js ` were updated so that errors returned are are converted to a liqualityErrorString for uniformity
2. Added additional translation method `$tle` primarily for translating liquality errors. This function might need to be updated so that the error suggestions are displayed in the desirable format (Currently, they are simply concatenated to the error cause)
3. Updated Vue files to use $tle function to translate liquality errors
4. `Timeline.vue` was updated to reflect the fact that item.error is now a liqualityErrorString (stringified JS Error) so $tle is used to translate it
5. Replaced Some custom errors with standard liquality errors.
6. Report errors instead of logging them to console.
7. Load Error translations into i18n instance

**General flow of Errors in this PR**
step 1: Errors(whether Liqualilty error or JS error) returned from wallet core to background.js and replyPermission.js is converted to a liqualityErrorString
step 2: liqualityErrorString is returned to Vue pages (either as 1 -  `message` field of JS Error, if from `background.js `2 - `error` field of `response` object, if from `replyPermission.js `or as error field of item as is the case with `Timeline.vue` file )
step 3:  $tle function of the Vue page is used to translate the liqualityErrorString and used to set the error variable of the vue page or anywhere else the error is been used. 

ps:  whenever there is a try/catch block, It cannot be certain whether the error caught is from `background.ts` or `replyPermission.js` or from erroneous codes in the Vue file,  hence the error is first converted to a liqualityErrorString before calling $tle. The conversion returns the same error if error is a liqualilty error, so conversion only happens otherwise). 

Link to Linear ticket:
https://linear.app/liquality/issue/EXT-36/error-and-exception-handling-error-service-qa-engineer


mpwanyi256 (changes requested):
@koderholic Well done, this is really great work. However, I have noticed a few things that we might look into before this PR is merged.

1. There are changes to the `What'sNewModal` we took out all firebase related stuff. So we might want to up date this PR with changes @Jennievon made in the PR she created.
2. Although we have covered atmost 90% of the errors in the application, we might need to add a catch for exceptions at the application root level using the config property on the vue instance. Kindly take a look at [The vue.js documentation](https://vuejs.org/api/application.html#app-config-errorhandler). This can be simply added in the main.js file where we are initalizing a new Vue.js instance to have somethi g like;

```
Vue.config.errorHandler = (err, instance, info) => {
  // handle error, e.g. report to a service
}
```

There is more information on the vue.js official site about what information is passed to the errorHandler function including which component and hook the error originated from.

mpwanyi256 (approved):
@koderholic CODE LGTM 🔥 

Approving this, we will cover the issues in my previous comments with @bradleySuira and @Jennievon thanks

PR #905, opened 2022-10-25
feat: Open customer questionnaire after user has deinstalled the chro…

beemi:

https://linear.app/liquality/issue/LIQ-1684/open-customer-questionnaire-after-user-has-deinstalled-the-chrome

### TODO

once we have a typesafe form we need to update this link in the background.js


kraikov (approved):

devaraj-s (approved):
LGTM - for my understanding what is the use of setting this URL

PR #919, opened 2022-11-08
LIQ-206 | NFT tab disappears after NFT count is zero fix

mpwanyi256:

### DIFF
--- a/src/components/WalletAccounts.vue
+++ b/src/components/WalletAccounts.vue
@@ -71,10 +71,13 @@
         </ListItem>
         <div class="account-assets" :class="{ active: shouldExpandAccount(account) }">
           <ListItem
-            v-if="account.nfts && account.nfts.length > 0 && !isAssetList"
+            v-if="account.nfts && !isAssetList"
             @item-selected="
               $router.push({
-                path: `/wallet/nfts/activity/${account.id}`
+                path: `/wallet/nfts/activity/${account.id}`,
+                query: {
+                  tab: getDefaultTab(account)
+                }
               })
             "
           >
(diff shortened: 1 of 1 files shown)


src/components/WalletAccounts.vue
```diff
@@ -71,10 +71,13 @@
...
         <div class="account-assets" :class="{ active: shouldExpandAccount(account) }">
           <ListItem
-            v-if="account.nfts && account.nfts.length > 0 && !isAssetList"
+            v-if="account.nfts && !isAssetList"
             @item-selected="
               $router.push({
-                path: `/wallet/nfts/activity/${account.id}`
+                path: `/wallet/nfts/activity/${account.id}`,
+                query: {
+                  tab: getDefaultTab(account)
```
Jennievon, line 79: @mpwanyi256 why are we adding this extra query? I don't see where it's being used

mpwanyi256: Screenshot 2022-11-08 at 14 48 12

@Jennievon We needed to direct the user with 0 NFT to the activity tab. So inside the `src/views/Wallet/NFTActivity.vue` component, there is already a check for that, I just needed to add a query with a default tab to switch the user to

src/components/WalletAccounts.vue
```diff
@@ -71,10 +71,13 @@
         </ListItem>
         <div class="account-assets" :class="{ active: shouldExpandAccount(account) }">
           <ListItem
-            v-if="account.nfts && account.nfts.length > 0 && !isAssetList"
+            v-if="account.nfts && !isAssetList"
```
Jennievon, line 74: @mpwanyi256 we need to also check that the NFT activity exists too.

mpwanyi256: @Jennievon Ohkay... thanks for the catch

mpwanyi256: @Jennievon Just pushed changes. Kindly take a look. thanks

Jennievon (approved):

bradleySuira (approved):
lgtm

PR #923, opened 2022-11-10
Fix cor 57 send transaction response from injection

bradleySuira:
# :books: Linear Ticket :books: #COR-57


mpwanyi256 (approved):
@bradleySuira CODE LGTM 🔥

PR #926, opened 2022-11-16
Enable decimal field customisation on pulling data from external token

mpwanyi256:


bradleySuira (approved):
LGTM

PR #927, opened 2022-11-21
LIQ-179 | Added user notification for existing custom tokens

mpwanyi256:

### DIFF
--- a/src/views/CustomToken.vue
+++ b/src/views/CustomToken.vue
@@ -61,7 +61,7 @@
               v-if="contractAddress && existingAsset"
               id="token_with_this_symbol_exits"
               class="text-danger form-text text-right"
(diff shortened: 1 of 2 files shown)


src/views/CustomToken.vue
```diff
@@ -172,6 +178,10 @@ export default {
       }
       return null
     },
+    tokenError() {
+      // TODO :: Need to add translations
+      return 'Already exists. Go to Token Assets.'
```
Jennievon, line 183: @mpwanyi256 you can add this to the English translation file and get it from there.

mpwanyi256: @Jennievon Yes I needed to do that. I just wans't sure how. But I have just updated the english translation file.

Thanks

PR #934, opened 2022-11-24
Error fixes while adding a custom token

mpwanyi256:

This PR Fiexes
- The 0 issue
- The errors shown when adding a custom token
- And the input issue where we needed to only accept numbers

### DIFF
--- a/src/views/CustomToken.vue
+++ b/src/views/CustomToken.vue
@@ -58,7 +58,7 @@
               required
             />
             <small
-              v-if="contractAddress && existingAsset"
+              v-if="contractAddress && existingAsset && !addingToken"
               id="token_with_this_symbol_exits"
               class="text-danger form-text text-right"
               >{{ $t('pages.customToken.tokenExists') }}</small
@@ -90,7 +90,7 @@
               :disabled="autofilled && !isSymbolEditable"
             />
             <small
-              v-if="symbol && symbolError"
+              v-if="symbol && symbolError && !addingToken"
               id="token_with_this_symbol_exits"
               class="text-danger form-text text-right"
               >{{ symbolError }}</small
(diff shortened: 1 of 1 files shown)


src/views/CustomToken.vue
```diff
@@ -99,9 +99,11 @@
           <div class="form-group">
             <label for="decimals">{{ $t('pages.customToken.decimals') }}</label>
             <input
-              type="text"
-              v-model="decimals"
+              type="number"
+              v-model.trim="decimals"
+              @keypress="isNumber($event)"
               class="form-control form-control-sm"
+              :maxlength="2"
```
Jennievon, line 106: @mpwanyi256 why are we using a max length of 2 here?

src/views/CustomToken.vue
```diff
@@ -305,6 +328,14 @@ export default {
       this.chainDropdownOpen = false
       this.resetFields()
       this.fetchToken()
+    },
+    isNumber(evt) {
+      const keysAllowed = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
+      const keyPressed = evt.key
+      // The Decimals field can only have 2 digits (0 - 99) and must be a number
+      if (!keysAllowed.includes(keyPressed) || this.decimals.length == 2) {
+        evt.preventDefault()
```
Jennievon, line 337: would be nice if we could move this to a util file so we could use it in other places

src/views/CustomToken.vue
mpwanyi256: @Jennievon This enables us to limit the value of decimal places to a maximum of 99.

src/views/CustomToken.vue
mpwanyi256: @Jennievon It would be, but this function implementation is specific to the fix here. It only validates for whole numbers and up to 2 digits. 

We could refactor this to better suite a general implementation but it would take more time looking into all possible use cases

Jennievon: okay then 👍

Jennievon (approved):

bradleySuira (approved):
LGTM

PR #940, opened 2022-12-01
fix: review validation to can send flag inside Send View

bradleySuira:
# :books: Linear Ticket :books: #ext-222


src/views/Send/SendNFT.vue
```diff
@@ -613,7 +615,16 @@ export default {
       this.sendFees = sendFees
     },
     updateSendFees: _.debounce(async function (amount) {
-      await this._updateSendFees(amount)
+      // In NFT screen the balance is equal to the native asset balance
+      console.log(
+        `on updateSendFees => amount: ${amount}, balance: ${this.balance}, asset: ${this.asset}, `
+      )
```
mpwanyi256, line 621: @bradleySuira I think we forgot to remove this console log

src/views/Swap/Swap.vue
```diff
@@ -211,7 +211,7 @@
               class="btn btn-primary btn-lg"
               id="swap_review_button"
               @click="review"
-              :disabled="!canSwap || cannotCoverNetworkFee"
+              :disabled="false"
```
mpwanyi256, line 214: @bradleySuira Do we need the prop if we are defaulting it to false?

bradleySuira: fixed man, thanks!

src/views/Send/SendNFT.vue
bradleySuira: we will keep this for debugging until we test this a lot

PR #953, opened 2022-12-07
Ext 234 only show self accounts tokens and non

bradleySuira:
# :books: Linear Ticket :books: #ext-234

This PR includes:
- Use Vue dev tools in development mode / check new task `dev:tools` inside package.json
- Added filter to use the same address for swap when is EVM accounts
- Added lazy loading routes


mpwanyi256 (approved):
@bradleySuira CODE LGTM 🔥 

Kudos on the routes import refactor

PR #955, opened 2022-12-09
Settings page and routes refactor

mpwanyi256:

This PR has the following changes;
- Refactored the Settings page to have children routes to persist the settings main navBar and footer
- Added a new link to Accounts settings aka Custom RPC settings


package.json
```diff
@@ -122,5 +123,11 @@
   "repository": {
     "type": "git",
     "url": "git@github.com:liquality/wallet.git"
+  },
+  "resolutions": {
```
mpwanyi256, line 127: @bradleySuira hahaha... Just a small not so we can remember to remove this before merging.

src/components/CustomRpcSettingsForm.vue
```diff
@@ -58,35 +49,69 @@ export default {
...
+        rpcUrl,
+        helperUrl,
+        custom } = network
+
+      const payload = {
+        network: this.activeNetwork,
+        walletId: this.activeWalletId,
+        chainId,
+        chanifyNetwork: {
+          ...network,
```
mpwanyi256, line 75: @bradleySuira If we are pulling all data from the network object here, we will need to remove lines 62 to 68

PR #956, opened 2022-12-09
fix: update error report config once state loads

koderholic:
This PR does the following

1. Updates error reporting config as soon as state is restored
2. Adds `Sentry` to `REPORT_TARGETS` in env

ps:
To unleash the power of sentry to help us debug, source map need to be uploaded for every release (https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/cli/)


bradleySuira (approved):
LGTM

PR #960, opened 2022-12-21
Set minimum amount that can be sent and prevent dust transactions

mpwanyi256:


bradleySuira (approved):
LGTM

PR #980, opened 2023-01-13
Ext 253/gas fee handler

mpwanyi256:

### DIFF
--- a/src/views/Send/Send.vue
+++ b/src/views/Send/Send.vue
@@ -9,16 +9,13 @@
         {{ $t('common.send') }}
       </NavBar>
       <InfoNotification v-if="nativeAssetRequired">
-        <EthRequiredMessage :account-id="account.id" :action="'send'" />
+        {{ $t('components.noFundsForNetworkFee.noEnoughAsset', { asset: assetChain }) }}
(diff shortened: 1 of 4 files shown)


src/views/Send/Send.vue
```diff
@@ -351,9 +357,23 @@ export default {
...
+    },
+    accountUrl() {
+      return `/accounts/${this.accountId}/${this.nativeAsset}/receive`
+    },
     account() {
       return this.accountItem(this.accountId)
     },
+    nativeAssetBalance() {
+      const balance = this.account.balances?.[this.nativeAsset] || 0
+      const pBalance = prettyBalance(balance, this.nativeAsset)
```
bradleySuira, line 371: @mpwanyi256 if you will need to convert the balance to BN, maybe we don't need to call `prettyBalance` just use the balance directly, `prettyBalance` is just to format and set the decimal places

bradleySuira (approved):
LGTM, just take in account the `prettyBalance` comment

src/views/Send/Send.vue
mpwanyi256: @bradleySuira Oh.. okay

bradleySuira (approved):

PR #982, opened 2023-01-17
Ext 256/custom rpc settins form fix

mpwanyi256:
This PR Fixes issues;

- Add Manage Accounts button inside settings (Should look like the Networks option)
- We should show the full address a user has entered instead of cutting it off.
- Double check to make sure that a user cannot edit the disabled fields


bradleySuira (approved):
LGTM

PR #992, opened 2023-02-06
Feature/enable all evm ledger

bradleySuira:

### DIFF
--- a/src/views/Accounts/HardwareWallet/HardwareWallet.vue
+++ b/src/views/Accounts/HardwareWallet/HardwareWallet.vue
@@ -26,7 +26,7 @@
       @on-cancel="cancel"
       @on-select-account="selectAccount"
     />
-    <Completed v-else />
+    <Completed v-else :selected-asset="selectedAsset" />
   </div>
 </template>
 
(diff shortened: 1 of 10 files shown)


src/views/Accounts/HardwareWallet/HardwareWallet.vue
```diff
@@ -189,42 +189,71 @@ export default {
...
+          publicKey,
+          chainCode
+        }
+
+        await this.createAccount({
+          network: this.activeNetwork,
+          walletId: this.activeWalletId,
+          account
+        })
+      }
```
mpwanyi256, line 222: @bradleySuira Well done man, Just a small change that we can make here to resole these promises more efficiently;

```
await Promise.all(accounts.map((key) => {
    const item = accounts[key]
        const { publicKey, chainCode, derivationPath } = item
        const index = item.index + 1
        const account = {
          name: `Ledger ${name} ${index}`,
          alias: '',
          chain,
          addresses: [item.account],
          assets,
          index: item.index,
          type: accountType,
          enabled: true,
          derivationPath,
          color: getNextAccountColor(chain, item.index),
          publicKey,
          chainCode
        }
        this.createAccount({
          network: this.activeNetwork,
          walletId: this.activeWalletId,
          account
        })
  }))
```

bradleySuira: we will iterate over the evm chains, taking your advice about the list or promises without any extra variable

